PR #24, opened 2026-03-19
Readme core structure

EvAvKein:
Just the empty sections we need to fill out, since we were talking about filling them out while working on the project. The sections' order isn't set in stone.


EvAvKein:
No idea why this includes the `backend compliation fix` change to `server.ts`, that should already be in main

EvAvKein:
Gonna merge into `dev` instead instead of `main`, since we've decided to funnel quick developments to there and thinking of `main` like 'production'

PR #28, opened 2026-03-24
Add configuration for importing files from an absolute path

EvAvKein:
This should let us avoid directory backtracking when importing into `.ts`/`.tsx` files, so the import routes would be simpler and less brittle.

Also, I included a small fix to our linting configuration. While working on this branch, I noticed that the linter (e.g. when we run `npm run SA`) would scan and complain about build-output files if those exist - it should no longer do that now.


EyzeCOLD (approved):
Good change. Very nice.

PR #36, opened 2026-03-26
Adds delete endpoint 

### LINKED ISSUE #32: File deletion

Omppu0:
* Added delete endpoint
* integrates it to the filebrowser
* fixed editor to not auto generate files

Closes #32

### DIFF
--- a/backend/src/endpoints/files.ts
+++ b/backend/src/endpoints/files.ts
@@ -154,6 +154,9 @@ const editFile = (app: Express, db: Pool) => {
 				parsedBody.data.content,
 				fileId.data,
 			]);
+			if (!result.rowCount || result.rowCount < 1) {
+				return res.status(403).send();
+			}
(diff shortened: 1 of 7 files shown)


backend/src/endpoints/files.ts
EvAvKein: `thingies` could have a more descriptive name?

frontend/src/dashboard/FileBrowser.page.tsx
```diff
@@ -15,11 +15,27 @@ function FileBrowserPage() {
 			});
 	}
 
+	async function handleDelete(id: string) {
```
EvAvKein: This function should probably be declared in `FIleList.tsx` since the delete buttons are there, and then pass down `refreshFileList` down to `<FileList>` the same way the refresh passed to `<FileUploader>`

EvAvKein (changes requested):
Small thing(ie)s to clean up, rest is good for this stage in development

EvAvKein (approved):
Good stuff

PR #37, opened 2026-03-27
Implements live (socket-based) collaborative file editing

### LINKED ISSUE #30: Implement collaborative live editing

EvAvKein:
Big pull request of a core feature, so we decided to review it all together in person soon.

Closes #30


backend/src/endpoints/collabSocket.ts
```diff
@@ -0,0 +1,263 @@
...
+	DocumentResponse,
+	ErrorResponse,
+	NameUpdateResponse,
+	SerializedUpdate,
+	UserFile,
+} from "#shared/src/types.js";
+
+const DATABASE_SAVE_DEBOUNCE_TIME = 1500;
+
+interface NameUpdate {
```
EyzeCOLD: Style-convention change interface -> type

backend/src/endpoints/collabSocket.ts
```diff
@@ -0,0 +1,263 @@
...
+	name: string;
+	version: number;
+}
+
+interface CollabSession {
```
EyzeCOLD: Style-convention change interface -> type

backend/src/endpoints/collabSocket.ts
```diff
@@ -0,0 +1,263 @@
...
+function drainPending<T>(pending: Array<(value: T) => void>, value: T): void {
+	while (pending.length) {
+		pending.pop()!(value);
+	}
+}
+
+function collabSocket(sockets: Server, db: Pool) {
+	const sessions = new Map<string, CollabSession>();
+
+	const getSession = async (fileId: string): Promise<CollabSession | undefined> => {
```
EyzeCOLD: Style-convention change arrow function to function function

backend/src/endpoints/collabSocket.ts
```diff
@@ -0,0 +1,263 @@
...
+				[fileId],
+			);
+			if (!result.rowCount) {
+				sessions.delete(fileId);
+				return undefined;
+			}
+			if (result.rowCount === 1) {
+				doc = Text.of((result.rows[0].content ?? "").split("\n"));
+				fileName = String(result.rows[0].name ?? fileName);
+			}
```
EyzeCOLD, line 66: We should probably have some sort of assertion for  `result.rowCount > 1`

backend/src/endpoints/collabSocket.ts
```diff
@@ -0,0 +1,263 @@
...
+		} catch (error) {
+			timestampedLog(`Error loading collab document ${fileId}: ${String(error)}`);
+		}
```
EyzeCOLD, line 71: Maybe some more drastic should happen on error? Part of the error handling and user feedback stuffs

backend/src/endpoints/collabSocket.ts
```diff
@@ -0,0 +1,263 @@
...
+		session.hasUnsavedChanges = false;
+		const content = session.doc.toString();
+
+		try {
+			const updateResult = await db.query("UPDATE files SET name = $1, content = $2 WHERE id = $3", [
+				session.fileName,
+				content,
+				fileId,
+			]);
+			if (updateResult.rowCount === 0) {
```
EyzeCOLD, line 113: Again could check if the rowCount is > 1.

backend/src/endpoints/collabSocket.ts
```diff
@@ -0,0 +1,263 @@
...
+				timestampedLog(`No file found to update for collab document ${fileId}`);
+				sessions.delete(fileId);
+			}
+		} catch (error) {
+			session.hasUnsavedChanges = true;
+			timestampedLog(`Error persisting collab document ${fileId}: ${String(error)}`);
+		}
+
+		if (session.hasUnsavedChanges && !session.dbSaveDebounceTimer) {
```
EyzeCOLD, line 132: Should this be inside the catch? What's up with the `!session.dbSaveDebounceTimer`, how can that happen?

backend/src/endpoints/collabSocket.ts
```diff
@@ -0,0 +1,263 @@
...
+			session.dbSaveDebounceTimer = setTimeout(() => flushSession(fileId), DATABASE_SAVE_DEBOUNCE_TIME);
+		}
+	};
+
+	const schedulePersist = (fileId: string, session: CollabSession) => {
```
EyzeCOLD: Could be renamed, maybe `scheduleFlush` or `toiletClock`

backend/src/endpoints/collabSocket.ts
```diff
@@ -0,0 +1,263 @@
...
+						break;
+					}
+					case "getInitialDocument": {
+						sendResponse({
+							version: session.updates.length,
+							doc: session.doc.toString(),
+						} satisfies DocumentResponse);
+						break;
+					}
+					case "getFileName": {
```
EyzeCOLD: Maybe rename?

backend/src/endpoints/collabSocket.ts
```diff
@@ -0,0 +1,263 @@
...
+							} satisfies NameUpdateResponse);
+						} else if (data.version === session.nameVersion) {
+							session.pendingName.push((nameUpdate: NameUpdate) => {
+								sendResponse(nameUpdate satisfies NameUpdateResponse);
+							});
+						} else {
+							sendResponse({
+								name: session.fileName,
+								version: session.nameVersion,
+							} satisfies NameUpdateResponse);
```
EyzeCOLD: There's two cases with the same outcome

backend/src/endpoints/collabSocket.ts
```diff
@@ -0,0 +1,263 @@
...
+						}
+						break;
+					}
+					case "pushFileName": {
+						const nextNameRaw = data.name;
+						if (typeof nextNameRaw !== "string") {
+							sendResponse({error: "Invalid file name"});
```
EyzeCOLD, line 222: More like undefined filename??

??
?

frontend/src/codeEditor/CodeEditor.tsx
```diff
@@ -1,7 +1,155 @@
...
+import langServer from "./langExtensions";
+import {CollabConnection, pushUpdates, pullUpdates, getInitialDocument} from "./collabClient";
+import styles from "./CodeEditor.module.css";
 
-export default function CodeEditor({value, onChange}: {value: string; onChange: ReactCodeMirrorProps["onChange"]}) {
-	return <CodeMirror theme={"dark"} extensions={[langs.markdown()]} value={value} onChange={onChange} />;
+const PUSH_MS_INTERVAL = 100;
+const PULL_MS_INTERVAL = 1000;
+
+interface CodeEditorProps {
```
EyzeCOLD: migrate to type

frontend/src/codeEditor/CodeEditor.tsx
```diff
@@ -1,7 +1,155 @@
...
+	const plugin = ViewPlugin.fromClass(LocalPeerPlugin);
+	return [collab({startVersion}), plugin];
+}
+
+export default function CodeEditor({fileId, onChange}: CodeEditorProps): JSX.Element {
+	const editorRef = useRef<HTMLDivElement>(null);
+	const connectionRef = useRef<CollabConnection | null>(null);
+	const [isLoading, setIsLoading] = useState(true);
+	const [error, setError] = useState<string | null>(null);
+	const onChangeRef = useRef(onChange);
```
EyzeCOLD: Type annotation missing?

frontend/src/codeEditor/CodeEditor.tsx
```diff
@@ -1,7 +1,155 @@
...
+		const editorElement = editorRef.current;
+		if (!editorElement) return;
+		let view: EditorView | null = null;
+
+		const initializeEditor = async (): Promise<void> => {
+			try {
+				connectionRef.current = new CollabConnection(fileId);
+				const {doc, version} = await getInitialDocument(connectionRef.current);
+				const state = EditorState.create({
+					doc: doc.toString(),
```
EyzeCOLD: `doc: doc` ?

frontend/src/codeEditor/Editor.page.tsx
```diff
@@ -1,67 +1,140 @@
...
 				navigate("/filebrowser");
 			});
+	}, [params.fileId]);
+
+	useEffect(() => {
+		if (!fileId) {
+			return;
+		}
+
+		const connection = new CollabConnection(fileId);
```
EyzeCOLD: Perhaps not used anymore, like the other filename thing you know......

EyzeCOLD (changes requested):
There was some things, but we did this in person so you know, you know.

backend/src/endpoints/collabSocket.ts
EvAvKein: Added an `else` statement that'll match rowCount above 1 (and also negative rowCounts, if that ever happens somehow)

backend/src/endpoints/collabSocket.ts
EvAvKein: Added session deletion. Can't really give more explicit user feedback from inside this function, and if we get into this catch block we likely have bigger issues than needing to inform the user

backend/src/endpoints/collabSocket.ts
EvAvKein: Added. I had to append a TypeScript assertion that it's not undefined/null (exclamation mark) at the end of the new rowCount check, for this query rowCount should always exist

backend/src/endpoints/collabSocket.ts
EvAvKein: Added a comment to be more explicit while why it's outside

backend/src/endpoints/collabSocket.ts
EvAvKein: Could do that, but maybe it's better to check "is this the correct type" (current check) rather than "is this not that one incorrect type"

🤷

EvAvKein:
Addressed all of the things we talked about in the review!

As per our idea in the group chat about not doing collaborative name-editing, I also removed the server-to-client file name updates. Users can now edit their own file from the editor page (using the same socket connection as the content editing) - currently still overwrites the shared file's name, but in the future it will just override the user's own personal file name (hence the name collab deletions).

EyzeCOLD (approved):
All issues rectified, good to go.

PR #47, opened 2026-04-04
FB - Frontend size and type validation for files

### LINKED ISSUE #46: FB - Frontend size and type validation for files

EyzeCOLD:
Filebrowser frontend blocks non-text filetypes and files over 1MB
* Some placeholder window.alert and window.confirm, can change them to
  less pop-uppy versions later
* Made things into tables
* Also played around with css
* The theme is kinda close to Tokyo Night... Could go all the way.

### DIFF
--- a/frontend/src/dashboard/FileUpload.tsx
+++ b/frontend/src/dashboard/FileUpload.tsx
@@ -13,7 +13,7 @@ function FileUploader({refreshFileList}: {refreshFileList: () => void}) {
 					console.error(`File '${f.name}' too large at ${f.size} (max. ${MAX_FILE_SIZE})`);
 					setFileUploads(null);
 					return;
-				} else if (f.type.slice(0, 5) !== "text/") {
+				} else if (!f.type.startsWith("text/")) {
 					window.alert(`File '${f.name}' is of unaccepted filetype ${f.type}`);
 					console.error(`File '${f.name}' is of unaccepted filetype ${f.type}`);
 					setFileUploads(null);
--- a/frontend/src/dashboard/FileUpload.tsx
+++ b/frontend/src/dashboard/FileUpload.tsx
@@ -3,8 +3,18 @@ import {useState, useRef} from "react";
 const FileUploader = ({refreshFileList}: {refreshFileList: () => void}) => {
 	const [fileUploads, setFileUploads] = useState<FileList | null>(null);
 	const fileInputRef = useRef<HTMLInputElement>(null);
+	const MAX_FILE_SIZE = 1000000; // 1meg
 
 	const handleFileChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+		if (e.target.files) {
+			for (const f of e.target.files) {
+				if (f.size > MAX_FILE_SIZE) {
+					console.error(`File '${f.name}' too large at ${f.size} (max. ${MAX_FILE_SIZE})`);
+					e.target.value = "";
+				}
+			}
+		}
+
 		if (e.target.files) {
 			setFileUploads(e.target.files);
 		}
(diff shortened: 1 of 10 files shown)


frontend/src/dashboard/FileUpload.tsx
```diff
@@ -39,23 +56,40 @@ const FileUploader = ({refreshFileList}: {refreshFileList: () => void}) => {
 	return (
 		<>
 			<div className="input-group">
-				<input ref={fileInputRef} id="file" type="file" multiple onChange={handleFileChange} />
+				<input
+					ref={fileInputRef}
+					style={{display: "none"}}
```
EvAvKein, line 61: Why is the input hidden?

EyzeCOLD: The "choose files" button apparently can't be styled, because it's somehow browser specific. So I'm doing this as a workaround.

frontend/src/dashboard/FileUpload.tsx
```diff
@@ -1,18 +1,35 @@
...
 
 	const handleFileChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+		if (e.target.files) {
+			for (const f of e.target.files) {
+				if (f.size > MAX_FILE_SIZE) {
+					window.alert(`File '${f.name}' too large at ${f.size} (max. ${MAX_FILE_SIZE})`);
+					console.error(`File '${f.name}' too large at ${f.size} (max. ${MAX_FILE_SIZE})`);
+					setFileUploads(null);
+					return;
+				} else if (f.type.slice(0, 5) !== "text/") {
```
Omppu0: string has ```.startsWith``` method so  ```f.type.startsWith("text/");``` would also work

EyzeCOLD: Yes. Fixed.

EyzeCOLD:
You can now run `npm run githook` to have the pre-push hook script setup for your local repo. To set up the script, the script is simply copied to .git/hooks/ folder, which isn't under version control. That means you have to do the setup yourself.

if you want to bypass the hook when pushing just use `git push --no-verify`.

frontend/src/dashboard/FileUpload.tsx
```diff
@@ -39,23 +56,40 @@ const FileUploader = ({refreshFileList}: {refreshFileList: () => void}) => {
...
+							{fileUploads &&
+								[...fileUploads].map((file) => (
+									<tr key={file.name}>
+										<td>{file.name}</td>
+										<td>{file.size} bytes</td>
+										<td>{file.type}</td>
+									</tr>
+								))}
+						</tbody>
+					</table>
```
EvAvKein, line 87: Doesn't affect the PR approval, but in the future should we maybe try to show the end-user only the info that's useful for them? The user likely knows their own files by name enough that we don't need to crowd the UI with the file size, and the technical name for the file type is probably not too helpful for them. A simple list of file names (file extensions included) should be enough, and would let the UI breath a bit.

frontend/src/index.css
```diff
@@ -1,9 +1,9 @@
 :root {
-	--text-primary: #fafafa;
-	--text-secondary: #999;
+	--text-primary: #dabfff;
+	--text-secondary: #7fdeff;
 
-	--background-primary: #171717;
-	--background-secondary: #232323;
+	--background-primary: #2c2a4a;
+	--background-secondary: #4f518c;
```
EvAvKein, line 6: Good colors ~

frontend/src/index.css
```diff
@@ -38,8 +38,28 @@ a {
...
+	padding-left: 8px;
+	padding-right: 8px;
+}
+
 button {
-	background-color: #000000;
+	background-color: var(--background-secondary);
+	padding: 4px 4px;
+	margin: 4px 2px;
 }
```
EvAvKein, line 63: Should probably add hover/focus styling. I was trying to tab-navigate on a page with a few a those buttons and had a hard type figuring out where my tab focus is.

Can do that in a later PR, I know I will 100% remember this.

EvAvKein (approved):

PR #52, opened 2026-04-07
Creates CodeMirror keybinds config, with Tab as a typeable character

EvAvKein:
I also took the liberty of rearranging the CodeEditor return values a bit, to minimize markup duplication.


Omppu0 (approved):
lgtm :)

PR #54, opened 2026-04-07
Add `package.json` audit commands and fix audit issues

EvAvKein:
There were a few minor, major, and critical vulnerabilities in dependencies. I figured adding the commands would make it easier for us to check our dependencies and update in future.

We could also allow GitHub's Dependabot to handle that stuff for us, if we decide we want less responsibility/control over this stuff.


Omppu0 (approved):
FINALLY

Issue #58, opened 2026-04-08
UM - Implement user data fetch endpoint


Sky11y:
Currently session only returns if it exists or not.
If needed it is easy to return more stuff from user.

Issue #57, opened 2026-04-08
UM - Forgotten password


Sky11y:
Implement user password resetting

Issue #35, opened 2026-03-26
Frontend - feedback about events


Omppu0:
* Show feedback on error
* Show feedback on success

EyzeCOLD:
🍞 TOAST 🍞

EvAvKein:
With #72 implemented, we just need to double-check that we're communicating errors and success consistently enough

PR #60, opened 2026-04-08
User management basics

### LINKED ISSUE #6: Implement user management

### LINKED ISSUE #16: Implement login page

### LINKED ISSUE #17: Implement account creation page

### LINKED ISSUE #18: Implement login endpoint & user data fetching

### LINKED ISSUE #19: Implement account creation backend

Sky11y:
PR for basic user management - simple frontend and data fetch from backend

## issues handled
#6 main issue - Implement user management
#16 - subissue - implement login page
#17 - subissue - implement account creation page
#18 - subissue - implement login endpoint and user data fetching
#19 - subissue - implement account creation backend

## Create tables in database
- users
- user\_sessions

## create endpoints
- /signup
- /login
- /dashboard
- /api/signup
- /api/session (login/logout/getsession)

## add npm modules
- argon2: password handling (hashing and verification)
- connect-pg-simple and express-session: handling user session
- express-rate-limit: limiting the number of request from same IP

## Other
- created backend middleware "requireAuth" for checking if user has an existing session
- created frontend function "getSession" for checking if user is authenticated

## Limitations
- Forgotten password not handled
    - Created new issue #57
- Get Session only returns valid session -> Could return other information if needed?
    - Created new issue #58
- Currently all the errors in forms launch window.alert. Change to toasts.
    - Already existing issue #35


backend/nginx/dev.nginx.conf
```diff
@@ -10,6 +10,8 @@ http {
   server {
     listen 80;
     listen [::]:80;
+    listen 8080;
+    listen [::]:8080;
```
EvAvKein: The compose.yaml connects the interal port 80 (the one listened to in this config) to the external 8080. This new listen at the Nginx container's internal 8080 isn't exposed or hooked up to anything.

backend/sql/sessions.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE IF NOT EXISTS "user_sessions" (
+    sid     varchar NOT NULL COLLATE "default",
+    sess    json NOT NULL,
+    expire  TIMESTAMP(6) NOT NULL
+)
+WITH (OIDS=FALSE);
+
+ALTER TABLE "user_sessions" ADD CONSTRAINT "session_pkey" PRIMARY KEY ("sid") NOT DEFERRABLE INITIALLY IMMEDIATE;
```
EvAvKein: Why is the table altered immediate after it's created? Can we not add this constraint in the table declaration like so?

````
CREATE TABLE IF NOT EXISTS "user_sessions" (
    sid     varchar NOT NULL COLLATE "default",
    sess    json NOT NULL,
    expire  TIMESTAMP(6) NOT NULL,
    CONSTRAINT "session_pkey" PRIMARY KEY ("sid")
);
```

backend/sql/sessions.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE IF NOT EXISTS "user_sessions" (
+    sid     varchar NOT NULL COLLATE "default",
+    sess    json NOT NULL,
+    expire  TIMESTAMP(6) NOT NULL
+)
+WITH (OIDS=FALSE);
```
EvAvKein: Seems like the `WITH (OIDS=FALSE)` is deprecated in Postgres versions newer than 12, and we have version 18. This `WITH` statement can be deleted.

backend/src/endpoints/users.ts
```diff
@@ -0,0 +1,108 @@
...
+import {z} from "zod";
+import rateLimit from "express-rate-limit";
+
+const limiter = rateLimit({
+	windowMs: 15 * 60 * 1000, // 15 min (how long to remember requests for)
+	limit: 5, // 5 attempts per 15 min
+	message: "Too many login attempts, please try again later.",
+});
+const signupUser = (app: Express, db: Pool) => {
+	app.post("/api/signup", async (req, res) => {
```
EvAvKein, line 15: Already mentioned this in person, but: To be more RESTful, this can be changed to post at `/api/user`. Not needed for this PR though, can be changed in future.

backend/src/endpoints/users.ts
```diff
@@ -0,0 +1,108 @@
...
+			});
+
+			await db.query("INSERT INTO users (username, email, hashed_password) VALUES ($1, $2, $3)", [
+				username,
+				email,
+				hash,
+			]);
+
+			console.log("Successfully created user");
+			res.status(201).json({success: "User Created"});
```
EvAvKein: No need to send the JSON, the frontend doesn't read it and doesn't seem like this is part of any existing pattern.

backend/src/endpoints/users.ts
```diff
@@ -0,0 +1,108 @@
...
+		} catch (err: unknown) {
+			if (err instanceof pgPromise.errors.QueryResultError && err.code === (23505 as number)) {
+				console.log("Client tried to create user with already existing name or email");
+				res.status(409).send("Username or email already in use");
+			} else if (err instanceof z.ZodError) {
+				const msg = err.issues[0].message;
+				res.status(401).json({error: msg});
```
EvAvKein: Isn't HTTP 401 more about unauthorized access to resources? I'd think that 400 is more appropriate for invalid input, even in sign-up.

backend/src/endpoints/users.ts
```diff
@@ -0,0 +1,108 @@
...
+			]);
+
+			console.log("Successfully created user");
+			res.status(201).json({success: "User Created"});
+		} catch (err: unknown) {
+			if (err instanceof pgPromise.errors.QueryResultError && err.code === (23505 as number)) {
+				console.log("Client tried to create user with already existing name or email");
+				res.status(409).send("Username or email already in use");
+			} else if (err instanceof z.ZodError) {
+				const msg = err.issues[0].message;
```
EvAvKein, line 37: We should probably make a ticket to improve users better about multiple errors in an input. This is currently just telling them about the first issue out of possibly multiple.

backend/src/endpoints/users.ts
```diff
@@ -0,0 +1,108 @@
...
+		}
+	});
+};
+
+const logoutUser = (app: Express) => {
+	app.delete("/api/session", (req, res) => {
+		req.session.destroy((err) => {
+			if (err) return res.status(500).json({error: "Logout failed"});
+			res.clearCookie("connect.sid");
+			res.json({message: "Logged out"});
```
EvAvKein: The frontend doesn't use the JSON, so that messag can be deleted. Since the frontend relies on the `ok`, we could explicitly send a 200 status code here instead.

backend/src/server.ts
```diff
@@ -6,13 +6,47 @@ import helmetSecurity from "helmet";
...
 import {collabSocket} from "./endpoints/collabSocket.js";
 
 const api = express();
 const server = createServer(api);
 const sockets = new Server(server, {cors: {origin: "*"}});
 collabSocket(sockets, postgres);
 
+/*=======USER SESSION==============*/
+
+declare module "express-session" {
```
EvAvKein: Can this declaration be removed from this file considering it's also in `express-session.d.ts`?

backend/src/types/express-session.d.ts
```diff
@@ -0,0 +1,8 @@
+import "express-session";
+
+declare module "express-session" {
+	interface SessionData {
+		userId: number;
+		username: string;
+	}
+}
```
EvAvKein: It's difficult (at least to me) to figure out where this is used. Can we document that?

backend/src/validation/schemas.ts
```diff
@@ -1,8 +1,28 @@
...
+const usernameSchema = z
+	.string()
+	.min(3, "Username has to be at least 3 characters long")
+	.max(20, "Username can't be over 20 characters long")
+	.regex(/^[a-zA-Z0-9_]+$/, "Username must contain only small/capital letter, numbers or underscore");
+
+export const UserSignupSchema = z.object({
+	username: usernameSchema,
+	email: z.email(),
+	password: passWordSchema,
```
EvAvKein: Is it just me or does the casing of `usernameSchema` and `passWordSchema` and `UserSignupSchema` seem a bit inconsistent?

backend/src/server.ts
```diff
@@ -6,13 +6,47 @@ import helmetSecurity from "helmet";
...
+		saveUninitialized: false,
+		cookie: {
+			maxAge: 24 * 60 * 60 * 1000, // 1 day
+			httpOnly: true,
+			//secure: true, //for https
+		},
+	}),
+);
+
+/*=========USER SESSION END==========*/
```
EvAvKein: Could the sessions setup be moved a different file/function? I'd think that would be the right approach when a segment is so big it needs comment delimiters, especially in a file like this which is supposed to be a high-level overview.

frontend/src/user/login.page.tsx
```diff
@@ -0,0 +1,82 @@
...
+			}
+
+			const response = await fetch("/api/session", {
+				method: "POST",
+				headers: {"Content-Type": "application/json"},
+				credentials: "include",
+				body: JSON.stringify({loginIdentifier, password: loginPassword}),
+			});
+			if (!response.ok) {
+				const msg = await response.text();
```
EvAvKein: The backend sends error response bodies in this endpoint as json objects with an `error` property. Shouldn't the message be the result of converting that response body from json and then getting the `error` property?

frontend/src/user/login.page.tsx
```diff
@@ -0,0 +1,82 @@
...
+			}
+			console.log("login successful");
+			navigate("/dashboard");
+		} catch (e) {
+			// TODO! Add the toast
+			window.alert(`Login failed: ${e}`);
+		}
+	};
+
+	//should we use maxlength for the input fields?
```
EvAvKein, line 46: Probably. Would make it harder for the user to make mistakes, which sounds like a good thing.

backend/src/endpoints/users.ts
```diff
@@ -0,0 +1,108 @@
...
+				email,
+				hash,
+			]);
+
+			console.log("Successfully created user");
+			res.status(201).json({success: "User Created"});
+		} catch (err: unknown) {
+			if (err instanceof pgPromise.errors.QueryResultError && err.code === (23505 as number)) {
+				console.log("Client tried to create user with already existing name or email");
+				res.status(409).send("Username or email already in use");
```
EvAvKein: Would be good to be consistent about the format we're sending error messages in - we're usually sending json, so should probably send this as json in with a `{error: string}` object.

frontend/src/user/signup.page.tsx
```diff
@@ -0,0 +1,96 @@
...
+			};
+
+			const response = await fetch("/api/signup", {
+				method: "POST",
+				headers: {"Content-Type": "application/json"},
+				credentials: "include",
+				body: JSON.stringify(newUser),
+			});
+			if (!response.ok) {
+				const msg = await response.text();
```
EvAvKein: The comment about login.page.tsx line 32 also applies here.

EvAvKein (changes requested):
One extra thing I didn't mention in the comments: A lot of function declarations (both frontend and backend) are assigning arrow functions to variables, which doesn't fit with our new TypeScript [conventions](https://github.com/EyzeCOLD/Transcendence/wiki/Conventions). Should change those declarations to `function` keyword declarations.

backend/nginx/dev.nginx.conf
Sky11y: Yeah forgot that there. Fixed.

backend/sql/sessions.sql
Sky11y: Yeah apparently legacy stuff that has no effect these days. I'll remove it.

backend/sql/sessions.sql
Sky11y: The recommended table creation is from the connect-pg-simple provider's page. Apparently legacy stuff. No problem to change like requested, so I'll do so.

backend/src/endpoints/users.ts
Sky11y: No. For a conflict the response code is 409.

backend/src/endpoints/users.ts
Sky11y: I'll change it to .send()

backend/src/endpoints/users.ts
EvAvKein: Oops this comment was on the wrong line, I meant line 38 not 35.

backend/src/endpoints/users.ts
Sky11y: Yes the status could should be there. I'll fix it and remove the text.

backend/src/validation/schemas.ts
Sky11y: Yes, the passWord should be password. I'll fix that. Other ways it seems to be okay. Maybe I could change the UserSignupSchema to just be SignupSchema

frontend/src/user/login.page.tsx
Sky11y: There's already a ticket for making the frontend input data validation better. So I think this can be resolved from this PR.

backend/src/endpoints/users.ts
Sky11y: Yes, makes sense. I'll change all to be {errror: string}.

frontend/src/user/login.page.tsx
Sky11y: Yes, You're correct. I'll fix that.

backend/src/endpoints/users.ts
Sky11y: Agreed that Eve does a ticket for this observation. Can be resolved.

backend/src/server.ts
Sky11y: Yes, makes sense. I created a file sessionConfig and moved session configuration related stuff there.

backend/src/server.ts
Sky11y: Yes, I created sessionConfig.ts file, which now has all the session related configuration

frontend/src/user/signup.page.tsx
Sky11y: Fixed this as well.

backend/src/endpoints/users.ts
EvAvKein: Created ticket #65

EvAvKein (approved):
Looks good, and all issues got addressed/resolved. Great work! 🙌

PR #63, opened 2026-04-09
Implements download button

### LINKED ISSUE #43: FB - Download and Export
Can a man just get a file on their disk?

The filename of the provided file should only be the basename. So if we have directories or something funky in there, it won't do path traversal for the client. Like if you use curl and then get your systemfiles overwritten with bullshit and we all go to jail.

EyzeCOLD:
* Some work done to tailwindify the style code
* Made a button component that works exactly like the html one, but has baked in style
* Removed the filetype checking on the frontend, since it was very dumb as it was

### DIFF
--- a/frontend/src/dashboard/FileUpload.tsx
+++ b/frontend/src/dashboard/FileUpload.tsx
@@ -1,4 +1,5 @@
 import {useState, useRef} from "react";
+import {Button} from "../components/Button";
 
 function FileUploader({refreshFileList}: {refreshFileList: () => void}) {
 	const [fileUploads, setFileUploads] = useState<FileList | null>(null);
@@ -11,11 +12,6 @@ function FileUploader({refreshFileList}: {refreshFileList: () => void}) {
 				if (f.size > MAX_FILE_SIZE) {
 					window.alert(`File '${f.name}' too large at ${f.size} (max. ${MAX_FILE_SIZE})`);
 					console.error(`File '${f.name}' too large at ${f.size} (max. ${MAX_FILE_SIZE})`);
-					setFileUploads(null);
-					return;
-				} else if (!f.type.startsWith("text/")) {
-					window.alert(`File '${f.name}' is of unaccepted filetype ${f.type}`);
-					console.error(`File '${f.name}' is of unaccepted filetype ${f.type}`);
 					setFileUploads(null);
 					return;
 				}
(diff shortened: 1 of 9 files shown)


frontend/src/dashboard/FileUpload.tsx
```diff
@@ -64,28 +60,24 @@ function FileUploader({refreshFileList}: {refreshFileList: () => void}) {
...
-										<td>{file.name}</td>
-										<td>{file.size} bytes</td>
-										<td>{file.type}</td>
+										<td>🗎 {file.name}</td>
 									</tr>
 								))}
 						</tbody>
 					</table>
-					<button onClick={handleUpload}>Submit</button>
+					<Button onClick={() => handleUpload()}>Submit</Button>
```
EvAvKein, line 80: The addition of the arrow function wrapper doesn't really do anything, but it matches the pattern of the other event callbacks so I think that's fine.

frontend/src/index.css
```diff
@@ -1,106 +1,78 @@
...
-	cursor: pointer;
-}
-button,
-a,
-summary,
-input,
-textarea {
-	font-size: inherit;
-	font-family: inherit;
-}
```
EvAvKein, line 86: Any particular reason for removing these cursor and font-family declarations? I'd say they're good to have, at least for a few of these elements.

frontend/src/index.css
```diff
@@ -1,106 +1,78 @@
 @import "tailwindcss";
 
-:root {
-	--text-primary: #dabfff;
-	--text-secondary: #7fdeff;
+@theme {
+	--color-foreground: #dabfff;
+	--color-accent: #7fdeff;
+	--color-canvas: #2c2a4a;
+	--color-surface: #4f518c;
```
EvAvKein, line 7: Are these names changeable? Feels like they could maybe be more descriptive (comparing to the previous ones)? Don't know if they only seem unintuitive to me

EvAvKein (commented):
Left two comments with questions

frontend/src/index.css
EyzeCOLD: They are. The thing is tailwind generates classes from them, so the "color-" part has to be there. The generated classes are gonna look something like `border-foregrund` or `bg-canvas`. I'd say let's adjust them further when we have some more of the tailwind classes in use to see if they read awkward.

frontend/src/index.css
EyzeCOLD: I tried to strip down the .css file so that it wouldn't have stuff that we don't actually use. It still has a lot. It's a work in progress.

EvAvKein: Makes sense. We can add these declarations back as/if needed

EvAvKein (approved):
Looks good!

PR #64, opened 2026-04-09
Small cleanups

EvAvKein:
- Removed placeholders from back when the project was created
- Renamed `backend/.env.template` to `backend/.env.example`, because the project's subject asks for the latter name


EyzeCOLD (approved):
🧹🧹🧹

PR #66, opened 2026-04-10
Adds filename collision checking

### LINKED ISSUE #50: FB - Duplicate filename checking on file creation and name change
Some kind of combination of a front end check and a more definitive back end check.
Check for filename collision on:
* Upload
* Multiple upload
* Filename update

EyzeCOLD:
* Added a nice link to the CodeEditor to jump back to files
* Code Editor filename changes are NOT handled yet, maybe in another PR? I'll need some consult from @EvAvKein most likely.

Closes #50

### DIFF
--- a/backend/src/endpoints/files.ts
+++ b/backend/src/endpoints/files.ts
@@ -22,7 +22,7 @@ function getFiles(app: Express, db: Pool) {
 			timestampedLog("Sent SELECT query to DB: all files");
 			const result = await db.query(`SELECT * FROM files`);
 			// console.log(result.rows);
-			return res.status(200).send(result.rows);
+			return res.status(200).json(result.rows);
 		} catch (error) {
 			console.log(`Error: ${error}`);
 			return res.status(500).send();
@@ -40,7 +40,7 @@ function getFileById(app: Express, db: Pool) {
 		// @TODO get auth stuff
 		const fileId = UserFileSchema.shape.id.safeParse(req.params.fileId);
 		if (!fileId.success) {
-			return res.status(400).send("Invalid file ID");
+			return res.status(400).json({error: "Invalid file ID"});
 		}
 
 		try {
@@ -52,7 +52,7 @@ function getFileById(app: Express, db: Pool) {
 				return res.status(403).send();
 			} else {
 				console.log(result.rows);
-				return res.status(200).send(result.rows[0]);
+				return res.status(200).json(result.rows[0]);
 			}
 		} catch (error) {
 			console.log(`Error: ${error}`);
@@ -68,7 +68,7 @@ function uploadNewFile(app: Express, db: Pool) {
 		const parsedBody = UserFileSchema.pick({name: true}).strict().safeParse(req.body);
 		if (!parsedBody.success) {
 			console.error("bad POST request", parsedBody.error);
-			return res.status(400).send("Bad request");
+			return res.status(400).json({error: "Bad request"});
 		}
 
 		try {
@@ -86,13 +86,13 @@ function uploadNewFile(app: Express, db: Pool) {
 				return res.status(403).send();
 			} else {
 				console.log(result.rows);
-				return res.status(201).send(result.rows[0]);
+				return res.status(201).json(result.rows[0]);
 			}
 		} catch (error: unknown) {
 			if (isDbError(error)) {
 				if (error.constraint === "files_name_key") {
-					console.error(`${error.detail}`);
-					return res.status(409).send(`${error.detail}`);
+					console.error({error: `${error.detail}`});
+					return res.status(409).json({error: `${error.detail}`});
 				}
 			}
 			console.error("Query failed:", error);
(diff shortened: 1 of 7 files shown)


backend/src/endpoints/files.ts
```diff
@@ -7,7 +7,12 @@ import multer from "multer";
 const storage = multer.memoryStorage();
 const upload = multer({storage: storage});
 
-const getFiles = (app: Express, db: Pool) => {
+// Type guard. Makes a type assertion for the error for TS.
+function isDbError(error: unknown): error is {code: string; detail?: string; constraint?: string} {
+	return typeof error === "object" && error !== null && "code" in error;
+}
```
EvAvKein, line 13: This seems like an excessive implementation for a function for this name. Is it not possible to simply return a check whether the given error is `instanceof` a DB error class?

backend/src/endpoints/files.ts
```diff
@@ -78,20 +82,26 @@ const uploadFile = (app: Express, db: Pool) => {
...
 			}
-		} catch (error) {
-			console.log("Query failed:", error);
+		} catch (error: unknown) {
+			if (isDbError(error)) {
+				if (error.constraint === "files_name_key") {
+					console.error(`${error.detail}`);
+					return res.status(409).send(`${error.detail}`);
+				}
+			}
```
EvAvKein, line 97: Changes based on the comment on line 13 might affect the implementation of this. The structuring of this error-handling seems unusual

frontend/src/codeEditor/Editor.page.tsx
```diff
@@ -66,6 +66,7 @@ export default function EditorPage() {
 					}}
 				/>
 			</label>
+			<Link to="/filebrowser">Fire Browser</Link>
```
EvAvKein: Fire browser? 🔥 🧑‍🚒 🧯 🚒 🚒

backend/src/endpoints/files.ts
```diff
@@ -78,20 +82,26 @@ const uploadFile = (app: Express, db: Pool) => {
...
 				console.log(result.rows);
 				return res.status(201).send(result.rows[0]);
 			}
-		} catch (error) {
-			console.log("Query failed:", error);
+		} catch (error: unknown) {
+			if (isDbError(error)) {
+				if (error.constraint === "files_name_key") {
+					console.error(`${error.detail}`);
+					return res.status(409).send(`${error.detail}`);
```
EvAvKein: This seems inconsistent with the other endpoints (at least the user-related ones), who usually send a `{error: string}` json. Should we maybe standardize to that obj structure that we can later potentially extend?

frontend/src/dashboard/FileUpload.tsx
```diff
@@ -42,6 +35,14 @@ function FileUploader({refreshFileList}: {refreshFileList: () => void}) {
 					body: formData,
 				});
 
+				if (result.status === 409) {
+					const text = await result.text();
```
EvAvKein: If the output from `files.ts` line 95 will change, should also change how you extract the error here.

backend/src/endpoints/files.ts
```diff
@@ -123,14 +133,20 @@ const uploadMultipleFiles = (app: Express, db: Pool) => {
 			const result = await db.query(query_string, argumentArray);
 			console.log("result rows", result.rows);
 			return res.status(201).send(result.rows);
-		} catch (error) {
+		} catch (error: unknown) {
+			if (isDbError(error)) {
+				if (error.constraint === "files_name_key") {
+					console.error(`${error.detail}`);
+					return res.status(409).send(`${error.detail}`);
```
EvAvKein: See the comment on line 95

backend/src/endpoints/files.ts
```diff
@@ -54,16 +59,15 @@ const getFileById = (app: Express, db: Pool) => {
...
-const uploadFile = (app: Express, db: Pool) => {
+function uploadNewFile(app: Express, db: Pool) {
 	app.post("/api/files", async (req, res) => {
 		// the front end could possibly check if a filename is valid (no repeats for a user/project)
 
 		const parsedBody = UserFileSchema.pick({name: true}).strict().safeParse(req.body);
 		if (!parsedBody.success) {
-			console.log("bad POST request", parsedBody.error);
+			console.error("bad POST request", parsedBody.error);
 			return res.status(400).send("Bad request");
```
EvAvKein: See comment on line 95

EvAvKein (changes requested):

backend/src/endpoints/files.ts
EyzeCOLD: I was thinking the same thing, but Claude claimed this

"Why are we defining it ourselves?
pg does have an internal error class (DatabaseError from the pg-protocol package that pg depends on), but it's not exported from pg itself in a straightforward way. You can get to it, but it's a bit awkward:
`typescriptimport { DatabaseError } from "pg-protocol";`
That works, and then you could do error instanceof DatabaseError instead of a manual type guard. Whether that's better is debatable — you're now depending on a sub-package of pg that isn't part of its public API, so it could break on a pg update without warning."

backend/src/endpoints/files.ts
EyzeCOLD: It's basically the exact same thing as in the user management side, except for what Claude was saying about the type not being a public api.

EyzeCOLD:
I did some refactoring, but I'm afraid a back and forth on Github is a really inefficient way to do this. We should probably just try to establish some sort of ENDPOINT FORMAT

EvAvKein (approved):
Nice!

PR #71, opened 2026-04-11
Add script to clone wiki and add wiki/ to .gitignore

EvAvKein:
Adds a `cloneWiki` script to the root `package.json` that clones the repo's wiki into `./wiki/`. The URL is derived from the git remote, so it'll work after a fork too. Also adds `wiki/` to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ dist
 .coverage
 .env
 *.swp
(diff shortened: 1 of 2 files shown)


Sky11y (approved):
Nice!

PR #73, opened 2026-04-14
Create toasts functionality

### LINKED ISSUE #72: Implement toasts on frontend

EvAvKein:
- Added Zustand as a dependency for frontend global state-management, which is used for toasts in this commit but in future could also be used for user state. We decided against the React built-in solution (Providers) since they'd reload more components unnecessarily.
- I heavily documented the WCAG compliance because there's some know-how to making toasts accessible (especially to the WCAG 2.1 AA standards from the subject's module).
- This replaces all the `window.alert` calls, but does not replace `window.confirm`. Making a `confirm` accessible toast alternative is a bit less trivial, so we can decide in future if that's really necessary.

Closes #72

### DIFF
--- a/frontend/src/components/toastStore.ts
+++ b/frontend/src/components/toastStore.ts
@@ -1,0 +1,39 @@
+import {create} from "zustand";
+
+export type ToastVariant = "error" | "success" | "info";
+
+type Toast = {
+	id: number;
+	message: string;
+	variant: ToastVariant;
+};
+
+type ToastStore = {
+	assertiveToasts: Toast[];
+	politeToasts: Toast[];
+	showToast: (variant: ToastVariant, message: string) => void;
+	dismiss: (variant: ToastVariant, id: number) => void;
+};
+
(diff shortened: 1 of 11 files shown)


frontend/src/components/toastStore.ts
```diff
@@ -0,0 +1,39 @@
...
+};
+
+let nextId = 0;
+
+export const useToastStore = create<ToastStore>((set) => ({
+	politeToasts: [],
+	assertiveToasts: [],
+	showToast: (variant: ToastVariant, message: string) => {
+		const id = nextId++;
+		if (variant == "error") set((state) => ({assertiveToasts: [...state.assertiveToasts, {id, message, variant}]}));
```
Sky11y: should be 3 equal signs on the condition?

Sky11y (commented):
Solid work. One typo fix and I'll accept the PR.

Sky11y (approved):
Good job!

PR #74, opened 2026-04-14
Session navigation fixes

### LINKED ISSUE #55: CE - Fix rapid rejoin error
Currently, `getInitialDocument` fails if you leave the code editor page and navigate back very quickly (e.g. browser history backward then forward), which results in an error page. Waiting a bit before navigating back avoids the problem, so this issue is specifically about rapid renavigation.

EvAvKein:
Closes #55, see that for details


EyzeCOLD (approved):
Approved.

Issue #31, opened 2026-03-25
Backend - User authentication checks for file access endpoints


Omppu0:

Omppu0:
Finished in #75

PR #75, opened 2026-04-14
User-based file ownership

### LINKED ISSUE #68: FB - File ownership
Currently, files have no association to a specific user — any authenticated user can access any file.

Each file should have an owning user (using the `user_id` foreign key from the database schema), with file endpoints scoped accordingly: the session user should be assigned as owner on upload, and fetch/modify/delete endpoints should be restricted to the owner.

This is also currently blocking #53.

EvAvKein:
Closes #68 & #31


EyzeCOLD (approved):
All good, except the navigation bug.

https://github.com/EyzeCOLD/DiffEd/issues/78

PR #77, opened 2026-04-15
Frontend fixes based on meeting notes

EvAvKein:
Based on our meeting notes today:
- Added password censoring to password input fields
- Extracted the toast colors to CSS variables
- Abstracted away the showToast extraction

I also added a layout rendering safeguard based on an earlier discovery today. This is not a full fix for the navigation issues we discovered, but should partially help.


frontend/src/user/signup.page.tsx
```diff
@@ -77,7 +77,12 @@ export default function SignupPage() {
 					<input placeholder="email" value={email} onChange={(e) => setUserEmail(e.target.value)} />
 				</div>
 				<div>
-					<input placeholder="password" value={password} onChange={(e) => setUserPassword(e.target.value)} />
+					<input
+						placeholder="password"
+						type="password"
```
EyzeCOLD, line 88: I assume this causes the password hiding? It should also be added to the "repeat password"

frontend/src/components/toastStore.ts
```diff
@@ -37,3 +37,7 @@ export const useToastStore = create<ToastStore>((set) => ({
 		}
 	},
 }));
+
+export function useShowToast() {
```
EyzeCOLD, line 41: There's a lot going in here that I don't get, but could we just export the straight up usable function showToast()? It feels still a bit awkward this way.

frontend/src/index.css
```diff
@@ -6,19 +6,12 @@
 	--color-canvas: #2c2a4a;
 	--color-surface: #4f518c;
 
-	--font-family-sans: Arial, sans-serif;
-}
+	--color-error-bg: #b91c1c;
+	--color-error-border: #ef4444;
+	--color-success-bg: #15803d;
+	--color-success-border: #22c55e;
```
EyzeCOLD: I suggest renaming these to not mention the element context, since these generate tailwind classes like `bg-error-bg` and `border-success-border`.

```
--color-error
--color-error-accent
--color-success
--color-success-accent
``` Maybe?

**Review by EyzeCOLD — CHANGES_REQUESTED** [review]
Some minor changes requested. Might also resolve the merge conflicts once you're at it.

File: frontend/src/user/signup.page.tsx (hunk 1 above)
Comment (EvAvKein) [reply frontend/src/user/signup.page.tsx:88]: Yes it does. Good catch, done!

File: frontend/src/components/toastStore.ts (hunk 2 above)
Comment (EvAvKein) [reply frontend/src/components/toastStore.ts:41]: Unfortunately that's not possible. `showToast` only works after the Zustand store gets constructed, it's dependant on initializing with `useToastStore`.

I could rewrite it like this if that'll help:
```
	const toastStore = useToastStore();
	return toastStore.showToast;
```

frontend/src/index.css
EvAvKein: Makes sense, can change that

EyzeCOLD (approved):
Great job, Eve!

PR #90, opened 2026-04-23
89 um endpoint refactoring

### LINKED ISSUE #89: UM: endpoint refactoring
Refactor the User Management endpoints to be in line with the pattern established in files.

EyzeCOLD:
* Migrates UM frontend to use apiFetch
* Refactored UM endpoints to follow the model established with the file endpoints

Closes #89

### DIFF
--- a/backend/src/endpoints/users.ts
+++ b/backend/src/endpoints/users.ts
@@ -1,13 +1,16 @@
-import {type Express} from "express";
+import type {Express, Request, Response} from "express";
 import {type Pool} from "pg";
-import pgPromise from "pg-promise";
 import argon2 from "argon2";
 import {SignupSchema, usernameSchema, emailSchema, passwordSchema} from "#/src/validation/schemas.js";
 import {z} from "zod";
-import {isDbError} from "#/src/utils.js";
+import {isDbError, isUniqueViolation} from "#/src/utils.js";
+import type {ApiResponse, SigningUser} from "#shared/src/types.js";
+import {timestampedLog} from "#/src/logging.js";
+import {requireAuth} from "#/src/middleware.js";
 
 function signupUser(app: Express, db: Pool) {
-	app.post("/api/user", async (req, res) => {
+	app.post("/api/user", async (req: Request, res: Response<ApiResponse<null>>) => {
+		timestampedLog(`REQUEST >>> ${req.method} ${req.url}`);
 		const {username, email, password} = req.body;
 
 		try {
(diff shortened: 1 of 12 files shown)


backend/src/endpoints/users.ts
```diff
@@ -16,140 +19,177 @@ function signupUser(app: Express, db: Pool) {
...
-			} else if (err instanceof z.ZodError) {
-				const msg = err.issues[0].message;
-				res.status(400).json({error: msg});
+			const query = "INSERT INTO users (username, email, hashed_password) VALUES ($1, $2, $3)";
+			const values = [username, email, hash];
+			timestampedLog(`DB QUERY >>> ${query}`);
+			timestampedLog(`DB VALUES >>> ${values}`);
+			await db.query(query, values);
+
+			res.status(200).json({ok: true, data: null});
```
Sky11y: I think this should stay as 201 'created'

frontend/src/user/login.page.tsx
```diff
@@ -23,26 +24,22 @@ export default function LoginPage() {
 	async function login(event: SubmitEvent<HTMLFormElement>) {
 		event.preventDefault();
 
-		try {
-			if (!loginIdentifier || !loginPassword) {
-				throw new Error("Please fill all the fields!");
-			}
+		if (!loginIdentifier || !loginPassword) {
+			throw new Error("Please fill all the fields!");
+		}
```
Sky11y, line 29: The try-catch block is removed, so this doesn't get thrown anywhere, i.e. the user doesn't get any feedback of empty fields.

frontend/src/user/userManagement.page.tsx
```diff
@@ -303,47 +303,39 @@ export default function UserManagementPage() {
...
 	) : (
 		<div>
 			<div>User Management</div>
 			<div>
-				<Username initialValue={user} onUpdate={handleUsernameUpdate} />
+				<Username initialValue={user} onUpdate={(newUsername: string) => setUser(newUsername)} />
 			</div>
 			<div>
-				<Email initialValue={email} onUpdate={handleEmailUpdate} />
+				<Email initialValue={email} onUpdate={(newEmail: string) => setEmail(newEmail)} />
```
Sky11y, line 338: Not 100 % sure, but I think these should be arrow functions, if they are inlined

backend/src/endpoints/users.ts
```diff
@@ -16,140 +19,177 @@ function signupUser(app: Express, db: Pool) {
...
+				const result = await db.query(query, [id]);
+
+				if (!result.rows.length) {
+					throw new Error("No query result");
+				}
 
-		const id = req.session.userId;
-		try {
-			const result = await db.query("SELECT username, email FROM users WHERE id = $1", [id]);
+				const user: Pick<SigningUser, "username" | "email"> = result.rows[0];
```
Sky11y: What's this pick thing?

Sky11y (changes requested):
Looks great! Few comments.

frontend/src/user/login.page.tsx
EyzeCOLD: Replaced the throw with a toast and a return

frontend/src/user/userManagement.page.tsx
EyzeCOLD: But they are arrow functions? What am I missing here?

backend/src/endpoints/users.ts
EyzeCOLD: Pick is TS magic that let's you create a temporary type by PICKING some fields out of an existing type.

I used this because I didn't want to create a new type for User that would just have username and email. However I had a chat with Eve and he said he would also need such a type, so we'll actually probably create that type and this ugly Pick thing will be replaced by that type.

EyzeCOLD: Actually I'll implement the type now.

backend/src/endpoints/users.ts
EyzeCOLD: Yes, I confused 201 created with 204 no content.

frontend/src/user/userManagement.page.tsx
Sky11y: Don't know what happened here. Everything's good!

backend/src/endpoints/users.ts
```diff
@@ -16,138 +19,171 @@ function signupUser(app: Express, db: Pool) {
...
+			} else {
+				timestampedLog(`ERROR <<< ${error}`);
+			}
+			return res.status(500).json({ok: false, error: "Internal server error"});
 		}
 	});
 }
 
+// Not a hundred percent on this type, didn't want to create a new one just for
+// this, but it felt bad using Object...
```
Sky11y: Comment outdated. Let's remove before finishing this PR.

backend/src/endpoints/users.ts
Sky11y: Thanks for the info!

Sky11y (approved):

backend/src/endpoints/users.ts
Sky11y: Fixed!

PR #91, opened 2026-04-28
Multi-user Collaboration & Misc

EvAvKein:
Main features of this PR:
- Multiple users can be in the same session, each selecting their own file
- File gets loaded and compared with CodeMirror's `unifiedMergeView` (git-like)
- Click file to create session, use new file-picker when you enter a session if you haven't selected one
- User can select which user they're comparing their file against, using the user buttons bar the top of the page

Misc changes:
- Created new `User` store on the frontend, in a dedicated stores directory (which now also has the toasts store).
- Added user data as a return value for the `POST /api/session` endpoint (useful for loading the frontend user store).
- Added ID as to the return value for the `GET /api/user` endpoint (not strictly necessary, but matches with the above-bullet's behavior of return user data inclusive of ID).
- Added a pixel of horizontal padding to `Input.tsx` to text input doesn't touch the border.
- Refactored `NewFile.tsx` to accept an optional `onCreated` hook, for use in `FilePicker.tsx`. Also, replace the button-as-form-label design with more frictionless and intuitive placeholder input text and submit button text.
- Moved all CodeMirror configs (keybinds, syntax lang, other extensions) into a dedicated file.
- Created `delay` util for frontend.

Todos (will discuss before converting to issues):
- Update CodeMirror color pallete from project colors, potentially with more intuitive/usage-appropriate diff colors.
- If the new `subscribe` socket event handling on the client seems to work well, consider subscription-based handling for `pullUpdates` and the like.
- Replace `RETRYABLE_INITIAL_DOC_MESSAGES` with something hopefully better than error-string-comparison.


Sky11y:
User can accept own changes.

Sky11y:
User could also have possibility to see user's own file without any diffs from other users in a session.

backend/src/endpoints/collabSocket.ts
```diff
@@ -62,199 +85,290 @@ function collabSocket(sockets: Server, db: Pool) {
...
-		if (session.hasUnsavedChanges) {
-			if (!session.dbSaveDebounceTimer) {
-				session.dbSaveDebounceTimer = setTimeout(() => {
-					void flushSession(fileId);
-				}, DATABASE_SAVE_DEBOUNCE_TIME);
-			}
-			return;
+		// If new changes arrived while we were awaiting the DB, reschedule.
+		if (slot.hasUnsavedChanges && !slot.dbSaveDebounceTimer) {
+			slot.dbSaveDebounceTimer = setTimeout(() => void flushOwnerSlot(slot), DATABASE_SAVE_DEBOUNCE_TIME);
```
Sky11y: Could call scheduleFlush instead of line 166

backend/src/endpoints/collabSocket.ts
```diff
@@ -62,199 +85,290 @@ function collabSocket(sockets: Server, db: Pool) {
...
-		session.dbSaveDebounceTimer = setTimeout(() => {
-			void flushSession(fileId);
-		}, DATABASE_SAVE_DEBOUNCE_TIME);
+		// Release any long-polling pullUpdates waiters so their clients fall through to the next cycle.
+		drainPending(slot.pending, []);
 	}
 
-	function attachSocketToSession(socketId: string, session: CollabSession) {
-		session.connectedSockets.add(socketId);
+	async function destroySession(shared: SharedSession): Promise<void> {
```
Sky11y: Maybe use other noun instead of Session (workspace/collabgroup etc.). There's already a userSession

backend/src/endpoints/collabSocket.ts
```diff
@@ -6,27 +6,46 @@ import type {Request, Response, NextFunction} from "express";
...
+	SessionInfo,
 } from "#shared/src/types.js";
 
 const DATABASE_SAVE_DEBOUNCE_TIME = 1500;
+// Gives users time to refresh, recover from a network blip, or navigate back before collab state is flushed
+const SESSION_DESTROY_GRACE_MS = 3000;
 
-type CollabSession = {
-	updates: Update[];
+type PerOwnerState = {
```
Sky11y: PerOwnerState could be global for all sessions instead of being a per session object. To handle different sessions changing the same file.

backend/src/endpoints/collabSocket.ts
```diff
@@ -62,199 +85,290 @@ function collabSocket(sockets: Server, db: Pool) {
...
-		session.hasUnsavedChanges = true;
-		if (session.dbSaveDebounceTimer) {
-			clearTimeout(session.dbSaveDebounceTimer);
+	async function evictSlot(slot: PerOwnerState): Promise<void> {
+		if (slot.dbSaveDebounceTimer) {
+			clearTimeout(slot.dbSaveDebounceTimer);
+			slot.dbSaveDebounceTimer = null;
+		}
+		if (slot.hasUnsavedChanges || slot.isFlushInProgress) {
+			await flushOwnerSlot(slot);
```
Sky11y: If at the end of flushOwnerSlot slot.hasUnsavedChanges == true another time to call flushOwnerSlot is placed. If that slot has already been deleted, seems like it will throw. Is this happening even possible?

backend/src/endpoints/collabSocket.ts
Sky11y: Could change the name of this type. Like just OwnerState or maybe even move away from the word Owner, because it's a little bit weird in this context. At the end of the day this represents the state of a single user or client in the collab/shared session.

Also if it gets moved away from this context, it should probably be even more descriptive like CollabDocument(State?) 

t. Jukka

backend/src/endpoints/collabSocket.ts
```diff
@@ -62,199 +85,290 @@ function collabSocket(sockets: Server, db: Pool) {
...
+			clearTimeout(slot.dbSaveDebounceTimer);
 		}
+		slot.dbSaveDebounceTimer = setTimeout(() => void flushOwnerSlot(slot), DATABASE_SAVE_DEBOUNCE_TIME);
 	}
 
-	function scheduleFlush(fileId: string, session: CollabSession) {
-		session.hasUnsavedChanges = true;
-		if (session.dbSaveDebounceTimer) {
-			clearTimeout(session.dbSaveDebounceTimer);
+	async function evictSlot(slot: PerOwnerState): Promise<void> {
```
Sky11y: evictSlot almost reads like a noun. Also you don't evict the slot, you evict the client or the document or the whatever. Like if I sit in a chair you don't kick out the chair, you kick me out. Right? evictClient? evictUser? clearSlot? nukeSlot?

t. Jukka

backend/src/server.ts
Sky11y: We need to change express back to "app" to be consistent. This starts to get confusing.

Sky11y (dismissed):

EvAvKein:
> User can accept own changes.

Fixed!

backend/src/endpoints/collabSocket.ts
EvAvKein: Done!

EvAvKein:
> User could also have possibility to see user's own file without any diffs from other users in a session.

Done, added "Solo" mode ~

backend/src/endpoints/collabSocket.ts
EvAvKein: Done! Renamed all collab sessions across the code to be called workspaces

EvAvKein:
Per our dislike/confusion with the bare CodeMirror styling yesterday, I made a theme using the project's colors. It even has borders around change sections, should make it easier to understand the scope of the Accept buttons!

backend/src/endpoints/collabSocket.ts
EvAvKein: Renamed things away from the excessive Owner usage, now the only place it's used in in a doc's ownerId (which seems reasonable?)

Renamed to LiveDocState, made more sense to me and close enough(?)

backend/src/endpoints/collabSocket.ts
EvAvKein: The refactoring to make doc states session-agnostic was going to remove all but one use of `evictSlot`, which I ended up just absorbing into the parent function that called it (new function named `releaseDocRef`). I also went through the rest of the code and removed all uses of the "slot" concept, which I also felt didn't quite fit.

backend/src/server.ts
EvAvKein: I dunno, to me `api` feels narrower and closer to the scope of that variable. If we wanna be consistent with the variable name the Express code samples give, sure we can do that (potentially in a different PR?)

backend/src/endpoints/collabSocket.ts
EvAvKein: Done!

backend/src/endpoints/collabSocket.ts
EvAvKein: I'm not sure which part of it would've thrown?

I'll talk about this in reference to the new code, since this has been moved a bit (relevant function is now `releaseDocRef`): When a doc gets "deleted" (e.g. `docs.delete`), that's just a deletion of the reference stored in the parent map, the doc variable still exists can be used for flushes as normal.

If the doc still has `hasUnsavedChanges` by this point in `flushDoc` (previously `flushOwnerState`), that's because the server received more changes while the current flush happened, and those changes' flush was terminated by `isFlushInProgress` being set to true in the ongoing flush call. It's either these new unflushed changes or the database failure in the catch block which would result in `hasUnsavedChanges` being true, and in both cases they need to be scheduled for a new flush (which will happen just fine with the doc object instance as it's still in these function calls' stack memory for the variable).

If this helps clarify that edge-case, let me know where/what code comment I can add for that. I already have some comments about the cases and the rest feels clear to myself, I'm not aware of the potential documentation gaps for this.

EvAvKein:
(Added Luka and Jukka because of the unfinished review thing from the group chat)

backend/src/server.ts
EyzeCOLD: The instance of Express being called app is kinda dumb convention to begin with, but whatever we call it, it would probably be best that it's always the same across the project. So in this case app.

Sky11y:
1. Currently the editor seems to delete own (not accepted) changes the user has made when accepting other user's changes.
I guess we need either add some save button or automatically save before accepting or something on those lines. I don't know if the saving would force the edit of other user's file as well, but the current behavior doesn't provide a very good user experience.

For now we can accept this behavior, so we can go forward with this PR, but an issue should be created for this.

2. Also I just noticed that deleting text messes up the editor. I painted some text and pressed delete -> I got 'accept' button but the co-worker didn't. Can't reproduce totally that anymore, but the 'accept' button still appears on my screen. -> Pressing the update button on own browser just recovers the deleted changes, i.e. no deletion happens.

backend/src/server.ts
EvAvKein: Pushed a commit to rename it back to `app`

frontend/src/layout/Layouts.tsx
```diff
@@ -17,20 +31,29 @@ export function PublicLayout() {
...
 	const navigate = useNavigate();
-	const [isLoggedIn, setIsLoggedIn] = useState(false);
-
-	useEffect(() => {
-		getSession().then(setIsLoggedIn);
-	}, []);
+	const storeUser = useCurrentUser();
+	const setUser = useSetUser();
+	const clearUser = useClearUser();
+	const [ready, setReady] = useState(!!storeUser);
```
Sky11y, line 37: Is the double !! some Typescript magic or typo?

frontend/src/stores/userStore.ts
Sky11y, line 1: Could you explain me what benefits does this store provide? It seems to work, so I'm not complaining. For me it just seems to complicate things.

frontend/src/fileBrowser/NewFile.tsx
```diff
@@ -1,50 +1,67 @@
...
-		console.log("New file created!");
+		if (onCreated) {
+			await onCreated(fileResult.data);
+			return;
+		}
+
+		const workspaceResult = await apiFetch<{workspaceId: string}>("/api/workspace", {
+			method: "POST",
+			body: JSON.stringify({fileId: fileResult.data}),
+			headers: [["Content-Type", "application/json"] as [string, string]],
```
Sky11y, line 38: No way Typescript can't implicitly understand that those are strings! LOL :D

ps. no need to fix anything, just an observation.

frontend/src/codeEditor/editorConfigs.ts
Sky11y, line 1: Let's view the editorConfigs/UI together later on. For now this seems quite working.

frontend/src/codeEditor/Editor.tsx
Sky11y, line 1: Is it possible to make it so that only the text field changes/loads when the diffed user is changed? i.e. could the input field and rename button persist, since changing the diffed user doesn't effect on those.

frontend/src/codeEditor/Editor.page.tsx
Sky11y, line 1: No toast for errors?

Sky11y (commented):
I didn't have time to go through the whole PR. Have to continue this some other time. Or we can share the load.

EvAvKein:
> 2. Also I just noticed that deleting text messes up the editor. I painted some text and pressed delete -> I got 'accept' button but the co-worker didn't. Can't reproduce totally that anymore, but the 'accept' button still appears on my screen. -> Pressing the update button on own browser just recovers the deleted changes, i.e. no deletion happens.

I went to reproduce this to try fixing the problem, and then realized that (if I correctly went through the steps) this is just potentially unintuitive behavior, but not a bug:

If two users have the same text in their docs, and user A deletes some text, user A will see an Accept block with that text because that's the diff - user B still has the text user A deleted, so user A is suggested to "Accept" the text user B has as an addition. That's also why user B doesn't have an Accept button - user B has (same as if it "added") text that user A does not have anymore.

From a technical perspective there's no bug here, but from a UX perspective there's an obvious issue. It's something that, at a glance, seems very difficult to solve though - we'd need to distinguish text that's intentionally deleted from text that the current user simply doesn't have. We can turn this into an issue if you want, but this is most likely not in-scope for our project deadline.

frontend/src/layout/Layouts.tsx
EvAvKein: JavaScript magic, but we could also do it in C/C++.

`!` is the "not" operator - turning the value into a boolean with the opposite truthiness. Double `!` just turns that opposite-truthiness boolean back to be the same truthiness as the original variable - converting the non-boolean value into a boolean

frontend/src/stores/userStore.ts
EvAvKein: It lets use have a global state for user data, so we don't have to contact the backend in every component we need their data e.g. username or id. With a store like this, we only need to fetch user data in the layout and populate there, so all other components use the store as a single source of truth.

A standard approach to reduce API calls for a resource without a store is to simply move the fetch up the call stack, but would create the issue of "prop-drilling" - a bad design practice where a lot of components that don't use some data require it as a paramater simply so they could pass it down to a deeply-nested component. Having it be "global" also avoids that, and it needs to be a "store" to be friendly to React's rerender logic.

frontend/src/fileBrowser/NewFile.tsx
EvAvKein: Oh it understands that those are strings! It's just strictly asking there for arrays which have exactly two strings, and without this `as` it treats that inner array as a variable-length string array ^^'

frontend/src/codeEditor/Editor.tsx
EvAvKein: Very possible, done!

frontend/src/codeEditor/Editor.page.tsx
EvAvKein: In this page I ended up showing errors with conditional HTML returns an `errorMessage` `useState`. Replacing those with toasts would leave the page pretty empty, and a toast would then leave the page completely once it disappears, hence the on-page error blocks.

backend/src/endpoints/collabSocket.ts
```diff
@@ -40,8 +66,15 @@ function drainPending<T>(pending: Array<(value: T) => void>, value: T): void {
 	}
 }
 
-function collabSocket(sockets: Server, db: Pool) {
-	const sessions = new Map<string, CollabSession>();
+function workspaceRoomName(workspaceId: string): string {
```
EyzeCOLD: Let's try to have function names be verbs. So `getWorkspaceName` or something. Probably also drop the room, since the workspace is the room, right?

backend/src/endpoints/collabSocket.ts
```diff
@@ -40,8 +66,15 @@ function drainPending<T>(pending: Array<(value: T) => void>, value: T): void {
...
+	return `workspace:${workspaceId}`;
+}
+
+export function collabSocket(sockets: Server, db: Pool): CollabSocketApi {
```
EyzeCOLD: Same here. If I understand correctly, this creates or initializes the collabSockets? So let's verb it up.

backend/src/endpoints/collabSocket.ts
```diff
@@ -62,199 +95,309 @@ function collabSocket(sockets: Server, db: Pool) {
 		}
 	});
 
-	async function getSession(fileId: string, userId: number): Promise<CollabSession | undefined> {
-		const session = sessions.get(fileId);
-		if (session) {
-			return session;
-		}
+	function buildWorkspaceInfo(shared: Workspace): WorkspaceInfo {
```
EyzeCOLD: Shared seems to be a leftover from the old naming, could rename to `workspace` or `ws`.

Same for other instances of a Workspace being passed as an argument.

backend/src/endpoints/collabSocket.ts
```diff
@@ -62,199 +95,309 @@ function collabSocket(sockets: Server, db: Pool) {
...
-				sessions.delete(fileId);
-				session.isFlushInProgress = false;
-				return;
+				timestampedLog(`Multiple files updated for collab doc ${doc.fileId}`);
 			}
 		} catch (error) {
-			session.hasUnsavedChanges = true;
-			timestampedLog(`Error persisting collab document ${fileId}: ${String(error)}`);
+			doc.hasUnsavedChanges = true;
+			timestampedLog(`Error persisting collab doc ${doc.fileId}: ${String(error)}`);
```
EyzeCOLD: I don't think you can use the verb "persist" in this way. Could it be reworded to "Error updating..." ?

backend/src/endpoints/collabSocket.ts
```diff
@@ -62,199 +95,309 @@ function collabSocket(sockets: Server, db: Pool) {
...
+			shared.destroyTimer = null;
+			// Someone reconnected during the grace window — skip destruction.
+			if (shared.connectedSockets.size > 0) return;
+			void destroyWorkspace(shared);
+		}, WORKSPACE_DESTROY_GRACE_MS);
 	}
 
-	function detachSocketFromSession(fileId: string, session: CollabSession, socketId: string) {
-		session.connectedSockets.delete(socketId);
+	function cancelWorkspaceDestroy(shared: Workspace): void {
```
EyzeCOLD: Is this redundant when the scheduled destruction already checks for the connections before killing the workspace?

backend/src/endpoints/collabSocket.ts
```diff
@@ -6,25 +6,51 @@ import type {Request, Response, NextFunction} from "express";
...
-type CollabSession = {
-	updates: Update[];
+type LiveDocState = {
+	ownerId: number;
+	username: string;
+	fileId: string;
+	fileName: string;
 	doc: Text;
+	updates: Update[];
 	pending: ((value: SerializedUpdate[]) => void)[];
```
EyzeCOLD: weird asymmetry, could be pendingUpdates or something

backend/src/endpoints/collabSocket.ts
```diff
@@ -62,199 +95,309 @@ function collabSocket(sockets: Server, db: Pool) {
...
+						shared.memberFiles.set(userId, data.fileId);
+						sendResponse(true);
+						try {
+							await broadcastMembers(shared);
+						} catch (err) {
+							timestampedLog(`Failed to broadcast members after pickFile: ${String(err)}`);
+						}
+						break;
+					}
+					case "leaveWorkspace": {
```
EyzeCOLD, line 363: It seems like the effects of this case and the socket.on("disconnect") could be generalized into a single function? DRY

backend/src/endpoints/collabSocket.ts
```diff
@@ -62,199 +95,309 @@ function collabSocket(sockets: Server, db: Pool) {
...
+		for (const existing of state.workspaces.values()) {
+			if (existing.memberFiles.get(userId) === fileId && existing.connectedSockets.size > 0) {
+				return existing.workspaceId;
+			}
 		}
 
-		if (session.isFlushInProgress || session.hasUnsavedChanges) {
-			void flushSession(fileId);
-			return;
+		if (!state.docs.has(fileId)) {
```
EyzeCOLD: The file loading code here and in the switch case's `pickFile` (lcollabSocket.ts:344) is kind of a DRY violation and could be generalized into a function.

EyzeCOLD (changes requested):
Some more review, I only had time to look at the first file. I'll get back to it later.

backend/src/endpoints/collabSocket.ts
EvAvKein: Renamed it to `buildWorkspaceName`, to align with the existing `buildWorkspaceInfo`

backend/src/endpoints/collabSocket.ts
EvAvKein: Renamed to `initCollabSocket`

backend/src/endpoints/collabSocket.ts
EvAvKein: Yup, done!

backend/src/endpoints/collabSocket.ts
EvAvKein: I think it's valid, but sure - switched it to "Error saving..."

backend/src/endpoints/collabSocket.ts
EvAvKein: To me it was intuitive because of the type data and how the vars are used, but no problem - renamed to `pendingUpdates`

backend/src/endpoints/collabSocket.ts
EvAvKein: Yup, leftover code. Deleted the function.

backend/src/endpoints/collabSocket.ts
EvAvKein: Did it with a few lines where it makes sense to me, but otherwise they have different functions - `leaveWorkspace` is immediate because it's intentional, and `disconnect` might be a network issue so it leaves time to reconnect (keeping the file in session in the meantime). If you really want, I can try to squeeze both cases into a shared function with branches set by boolean parameters, but that's not gonna be very readable at the call sites.

backend/src/endpoints/collabSocket.ts
EvAvKein: Valid, moved that external file-loading logic into the adjacent `loadOwnerDoc`s (and renamed it to `loadDoc` to further get rid of excessive "owner" wording)

frontend/src/layout/Layouts.tsx
EyzeCOLD: It's a bit of a hack, but it's also such a well-known oldschool hack, so maybe it's okay

Sky11y:
> > 2. Also I just noticed that deleting text messes up the editor. I painted some text and pressed delete -> I got 'accept' button but the co-worker didn't. Can't reproduce totally that anymore, but the 'accept' button still appears on my screen. -> Pressing the update button on own browser just recovers the deleted changes, i.e. no deletion happens.
> 
> I went to reproduce this to try fixing the problem, and then realized that (if I correctly went through the steps) this is just potentially unintuitive behavior, but not a bug:
> 
> If two users have the same text in their docs, and user A deletes some text, user A will see an Accept block with that text because that's the diff - user B still has the text user A deleted, so user A is suggested to "Accept" the text user B has as an addition. That's also why user B doesn't have an Accept button - user B has (same as if it "added") text that user A does not have anymore.
> 
> From a technical perspective there's no bug here, but from a UX perspective there's an obvious issue. It's something that, at a glance, seems very difficult to solve though - we'd need to distinguish text that's intentionally deleted from text that the current user simply doesn't have. We can turn this into an issue if you want, but this is most likely not in-scope for our project deadline.

Oh yeah, that's true. Sounds complicated to fix. So let's accept this behavior.

EyzeCOLD (approved):
Phew. Great work, but let's keep them smaller in the future. Like way smaller.

PR #92, opened 2026-04-30
Db separating queries

Sky11y:
Pulled out all the postgres queries from backend users.ts and sessions.ts to a new file called /queries/users.ts.
This PR does not close any tickets/issues, but makes user related database queries more cleaner.


backend/src/endpoints/users.ts
```diff
@@ -1,37 +1,34 @@
 import type {Express, Request, Response} from "express";
-import {type Pool} from "pg";
 import argon2 from "argon2";
 import {SignupSchema, usernameSchema, emailSchema, passwordSchema} from "#/src/validation/schemas.js";
-import {z} from "zod";
 import {isDbError, isUniqueViolation} from "#/src/utils.js";
 import type {ApiResponse, User} from "#shared/src/types.js";
 import {timestampedLog} from "#/src/logging.js";
 import {requireAuth} from "#/src/middleware.js";
+import * as userQueryService from "#/src/queries/users.js";
```
EyzeCOLD: You can just do `import userQueryService from "#/src/queries/users.js";` and then in users.ts have `export default { ... }`, would look hotter.

EyzeCOLD (dismissed):
There's a small comment, resolve that and you can merge.

backend/src/endpoints/users.ts
Sky11y: Done! Also needed to fix the sessions.ts import.

EyzeCOLD (approved):

PR #93, opened 2026-05-06
48 fb   backend filetype and size capping

### LINKED ISSUE #44: FB - Filetype validation, size capping
Right now binary files and big files are not handled well. Let's change that.

### LINKED ISSUE #48: FB - Backend filetype and size capping

### LINKED ISSUE #84: FB - Add ability to remove files from upload list

Omppu0:
* Merged file endpoints that have similar functionality.
* Fixed parts of the frontend that depended on those endpoints.  
* Added filetype and size checking to back and frontend.
* Extended file upload functionality to have better error messaging. Frontend filters out invalid files when appending to the form. Now it wont remove the files from the form on error unless the error is unrecoverable.
* Adds ability to remove files from the form or append to it before a successful submit.

* Closes #44 
* Closes #48  
* Closes #84


EyzeCOLD (changes requested):
Yes, yes, well done...however

- backend/src/endpoints/files.ts
    - :116 - This is probably just debug stuff? If you think we need to keep the log, please use timestapedLog()
    - :126 - Let's try to avoid magic numbers, maybe make a function or a constant or something for this?
- frontend/src/fileBrowser/FileBrowser.page.tsx
    - :23 - Can't you instead make some magic runes to quell this specific ESLint thing? The 1:1 duplication is horrible.
- frontend/src/fileBrowser/FileList.tsx
    - :4 - Weird whitespace but okay
- frontend/src/fileBrowser/FileUpload.tsx
    - :31- Doesn't this allow you to add multiple files with same names as long as they have different sizes? Wouldn't make much sense since you can't upload them anyway
    - :35 - Should probably be "error" and not info. We could also already start taking out all the console.errors() from frontend.
    - :64 - Agreed, but we should still probably show the toast, especially when "greying out" the button isn't implemented.
- shared/src/fileTypeCheck.ts 
    - Could be renamed to fileCheck or fileValidation or something. Also the function doesn't return a boolean so it should be a verb.

EyzeCOLD (approved):

PR #96, opened 2026-05-06
FB Pagination, sorting and filtering

### LINKED ISSUE #95: FB: Pagination, Sort, Search

EyzeCOLD:
* Paginator
* Sort and filter
* FilePicker now uses FileBrowser, so Paginator, sort and filter also in FilePicker

Closes #95


EvAvKein:
Stuff to fix from checking them out in-person:
- Put file list filter into hide condition when no files
- Remove page interface buttons (Paginator) when single page
- Remove "Page X out of Y" text

frontend/src/fileBrowser/useFileList.ts
```diff
@@ -0,0 +1,49 @@
...
+import {apiFetch} from "#/src/utils";
+import {useShowToast} from "#/src/stores/toastStore";
+
+function useFileList() {
+	const [fileList, setFileList] = useState<UserFile[] | null>(null);
+	const [filter, setFilter] = useState("");
+	const [page, setPage] = useState(0);
+	const [sortDescending, setSortDescending] = useState(true);
+	const showToast = useShowToast();
+	const FILES_PER_PAGE = 10;
```
EvAvKein: Can put this outside of the component render process, top of the file

EyzeCOLD: Donatello.

frontend/src/components/Button.tsx
EvAvKein, line 1: I'm not sure why `disabled` has a separate return statement? To reduce markup duplication, couldn't it specifically affect the relevant classes (and probably set the `disabled` HTML button attribute for accessibility)? The earlier return statement also omits the props spreading - if a developer is not conscious of this particular implementation detail, it could accidentally introduce future bugs

frontend/src/fileBrowser/Paginator.tsx
```diff
@@ -0,0 +1,52 @@
...
+			})}
+			<Button
+				onClick={() => onPageChange(Math.min(currentPage + 1, totalPages - 1))}
+				disabled={currentPage === totalPages - 1}
+			>
+				»
+			</Button>
+		</div>
+	);
+}
```
EvAvKein: Some accessibility improvements:
- Replace the `<div>` wrapper with something like `<nav aria-label="Pagination">`: Screenreaders show the landmark role that the nav implicitly has, which lets the screenreader user go directly to pagination.
- Add aria-labels (something like `aria-label="Previous page"` and `aria-label="Next page"`) to the « and » buttons: The raw symbols are read by the glyph name or skipped, the usage is not intuitive for screenreader users.
- Add `aria-current="page"` to the current page button: The style classes aren't enough for screenreader users. That will re-announce the current page when the user navigates to the next page.
- Add more descriptive `aria-label="more pages"` tags to the `<span>...</span>` elements.

Also, the elements returned by `.map` here also don't have a `key` attribute, which they probably should.

EvAvKein:
Commented about a few things worth fixing, I'll continue my review later today

frontend/src/fileBrowser/FileBrowser.tsx
```diff
@@ -0,0 +1,53 @@
...
+		sortDescending,
+		setSortDescending,
+		refreshFileList,
+		totalFiles,
+	} = useFileList();
+
+	return (
+		<>
+			{totalFiles > 0 ? (
+				<div>
```
EvAvKein: This div wrapper is not doing anything. Removable?

frontend/src/fileBrowser/FileBrowser.tsx
EvAvKein, line 1: The structure of the new file stuff, or at least the naming, is a bit hard to follow.

`useFileList` is not used by `FileList.tsx`, but `FileBrowser.tsx`?

The component `FileBrowser.tsx` has the same name as `FileBrowser.page.tsx`, but is a reusable component (reused in `FilePicker.tsx`) - should the reusable have a different name than the page?

The `Paginator` sounds like the thing that would *do* (or at least include the whole of) pagination, but it only renders the buttons based on the parameters - should it have a more component-y name than an action-y name?

frontend/src/fileBrowser/useFileBrowser.ts
```diff
@@ -0,0 +1,52 @@
...
+		const response: ApiResponse<UserFile[]> = await apiFetch("/api/files");
+		if (response.ok) {
+			setFileList(response.data);
+		} else {
+			console.error(response.error);
+			showToast("error", `${response.error}`);
+		}
+	}
+
+	useEffect(() => void refreshFileList(), []);
```
EvAvKein, line 32: We already talked about the `eslint/react-hooks` IDE warning for this in person... Since this PR is already being held up by my requesting changes, should maybe try to figure out which way we'd be able okay with resolving it?

frontend/src/fileBrowser/useFileList.ts
```diff
@@ -0,0 +1,52 @@
...
+			console.error(response.error);
+			showToast("error", `${response.error}`);
+		}
+	}
+
+	useEffect(() => void refreshFileList(), []);
+
+	const totalFiles = fileList ? fileList.length : 0;
+	const paginated = processed.slice(page * FILES_PER_PAGE, (page + 1) * FILES_PER_PAGE);
+	const totalPages = processed ? Math.ceil(processed.length / FILES_PER_PAGE) : 0;
```
EvAvKein: The type of `processed` is `UserFile[]`, so this ternary is never false

frontend/src/fileBrowser/Paginator.tsx
```diff
@@ -0,0 +1,52 @@
...
+	return (
+		<div>
+			<Button onClick={() => onPageChange(Math.max(0, currentPage - 1))} disabled={currentPage === 0}>
+				«
+			</Button>
+			{pageNumbers.map((p) => {
+				if (p === "...") {
+					return <span>...</span>;
+				} else if (p === currentPage + 1) {
+					return <Button className="underline underline-offset-2 font-bold">{p}</Button>;
```
EvAvKein: A button without an `onClick` feels a bit misleading, could we make this not a button? Or maybe keep it a button, but make the button `disabled`? Talking about the HTML attribute, but potentially also our component parameter if that'll still be a thing after the changes requested

EvAvKein (changes requested):

frontend/src/components/Button.tsx
EyzeCOLD: Now it has just one return with a function that picks from two constant style strings for the classNames

frontend/src/fileBrowser/Paginator.tsx
EyzeCOLD: There was some wild shit about using `aria-current="page"` with `disabled` according to Claude, so I settled on a kind of a compromise. It is a button, but for sighted users it's very clear that the button doesn't do anything.

frontend/src/fileBrowser/FileBrowser.tsx
EyzeCOLD: 🧹🧹

frontend/src/fileBrowser/FileBrowser.tsx
EyzeCOLD: I renamed `Paginator` and `useFileList` (that was really bad), but don't like renaming `FileBrowser.tsx`. I'm open to renaming `FileBrowser.page.tsx`, but I still hesitate to do that because then we should rename the route and the folder and all that.

Let me know what you think

frontend/src/fileBrowser/useFileBrowser.ts
EyzeCOLD: I emloyed some react tech, let's see if `useCallback()` fixes it for you

frontend/src/fileBrowser/useFileList.ts
EyzeCOLD: Oh yeah, it always returns an array. Fixed.

frontend/src/fileBrowser/Paginator.tsx
EyzeCOLD: Like I alluded to before, apparently some screen readers will ignore buttons that are disabled, which is a no-no for an `aria-current="page"`. I made a compromise that to me feels like a good balance between the screen reader getting all necessary info and sighted users not being confused either.

frontend/src/fileBrowser/useFileBrowser.ts
EyzeCOLD: Okay, I tried with Zed and this didn't help. I added a rule to our frontends eslint config. That should fix it for the CI and the editor all at once.

frontend/src/fileBrowser/NewFile.tsx
```diff
@@ -1,18 +1,15 @@
...
 import {apiFetch} from "#/src/utils.ts";
 
 type NewFileProps = {
-	// When provided, called with the new file's ID instead of creating a workspace and navigating
-	onCreated?: (fileId: string) => Promise<void>;
+	onFileSelect: (fileId: string) => void;
 };
 
-function NewFile({onCreated}: NewFileProps): JSX.Element {
+function NewFile({onFileSelect}: NewFileProps): JSX.Element {
```
EvAvKein: Why is it called `onFileSelect` now? Seems less descriptive than the previous name

frontend/src/fileBrowser/NewFile.tsx
```diff
@@ -1,18 +1,15 @@
...
 	const [newFilename, setNewFilename] = useState<string>();
```
EvAvKein: Not related to this PR, but might as well fix it while you're editing the file again: Can we add an empty string as the initial useState value? Right now it's undefined, which resolves to the file name "undefined" if the user just clicks the Create button straight away

frontend/src/fileBrowser/FileBrowser.tsx
EvAvKein: The new names make good enough sense. Regarding "PageSelector" I would've avoided the second word because of CSS selectors, but not a big deal. There's potential to simplify the route to `/files` and page to `files.page.tsx`, but that's not necessarily better

EvAvKein (changes requested):
Just two small things in `NewFile.tsx` and we're finished 🙏

frontend/src/fileBrowser/NewFile.tsx
EyzeCOLD: Oh yeah, fixed

frontend/src/fileBrowser/NewFile.tsx
EyzeCOLD: I wanted to use the same name for both NewFile and FileList, but you're right, there's no reason for it to be more descriptive.

EyzeCOLD:
Small fixes pushed

EvAvKein (approved):
Good stuff! There were some styling/design things, but we said we'll get to those in the coming days

PR #97, opened 2026-05-07
FilePicker-adjacent UX improvements

EvAvKein:
- Fixed file-picker showing on momentary disconnect (e.g. page refresh)
  - I did this by expanding the workspace-destroy grace period to affect member disconnects.
- Introduced button in editor page to send oneself back to the file picker.
  - This works by, when a new file is selected, rapidly disconnecting member with the previous file and reconnecting it with a new file. This quick reconnect is not very noticeable for the repicker, but potentially a bit jarring for the collaborators. This is maybe a bit hacky and could be replaced with a more seamless file-swap, but this soluition is more reviewable/straight-forward and probably fine considering our time budget.


EyzeCOLD (approved):
All looks good

PR #112, opened 2026-05-07
OAuth with GitHub

### LINKED ISSUE #87: Implement OAuth for github login

EvAvKein:
- Created a module for env stuff so we error better on missing env
- Small adjusments per recent team meeting:
  - Update favicon
  - Delete `pg-promise` unused dependency
- GitHub OAuth with PassportJS

Closes #87


backend/src/endpoints/oauth.ts
```diff
@@ -0,0 +1,190 @@
...
+import {SESSION_SECRET} from "#/src/env.js";
+import {timestampedLog} from "#/src/logging.js";
+import {isDbError, isUniqueViolation} from "#/src/utils.js";
+import {usernameSchema} from "#/src/validation/schemas.js";
+import type {ApiResponse, User, PendingGithubPayload} from "#shared/src/types.js";
+import {requireAuth} from "#/src/middleware.js";
+import userQueryService from "#/src/queries/users.js";
+
+function signToken(payload: PendingGithubPayload): string {
+	const data = Buffer.from(JSON.stringify(payload)).toString("base64url");
```
EyzeCOLD, line 17: _"While the Buffer class is available within the global scope, it is still recommended to explicitly reference it via an import or require statement."_

https://nodejs.org/api/buffer.html#buffer

backend/src/endpoints/oauth.ts
```diff
@@ -0,0 +1,190 @@
...
+	const dotIndex = token.lastIndexOf(".");
+	if (dotIndex === -1) return null;
+
+	const data = token.slice(0, dotIndex);
+	const signature = token.slice(dotIndex + 1);
+	const expectedSig = crypto.createHmac("sha256", SESSION_SECRET).update(data).digest("base64url");
+
+	if (!crypto.timingSafeEqual(Buffer.from(signature), Buffer.from(expectedSig))) return null;
+
+	const payload: PendingGithubPayload = JSON.parse(Buffer.from(data, "base64url").toString());
```
EyzeCOLD: Not sure if this and the `timingSafeEqual()` above should be in a try-catch, since both of them could throw?

backend/src/endpoints/oauth.ts
```diff
@@ -0,0 +1,190 @@
...
+function setGithubUsername(app: Express) {
+	app.post("/api/auth/github/username", async (req: Request, res: Response<ApiResponse<User>>) => {
+		timestampedLog(`REQUEST >>> ${req.method} ${req.url}`);
+
+		const {token, username} = req.body;
+		if (!token || !username) {
+			return res.status(400).json({ok: false, error: "Missing token or username"});
+		}
+
+		const payload = verifyToken(token);
```
EyzeCOLD, line 137: I got obsessed with verifyToken so I tried looking at it with Claude. There's some curls that can make stuff throw that's not caught by us.

Claude gave me these lines: 
- Makes `indexOf` throw: ` curl -X POST http://localhost:8080/api/auth/github/username   -H "Content-Type: application/json"   -d '{"token": 123, "username": "test"}'`
- Makes `timingSafeEquals()` throw: `curl -X POST http://localhost:8080/api/auth/github/username   -H "Content-Type: application/json"   -d '{"token": "eyJnaXRodWJJZCI6IjEyMyJ9.shortsi g", "username": "test"}'`

Seems that it is still handled by Express, but it probably should be error handled by our code? Nothing actually crashes though.

backend/src/endpoints/oauth.ts
```diff
@@ -0,0 +1,190 @@
...
+
+	const payload: PendingGithubPayload = JSON.parse(Buffer.from(data, "base64url").toString());
+	if (Date.now() > payload.expiration) return null;
+
+	return payload;
+}
+
+function githubAuthStart(app: Express) {
+	app.get("/api/auth/github", (req: Request, res: Response, next) => {
+		const action = (req.query.action as string) || "login";
```
EyzeCOLD: Should we rather use an allowlist instead of passing the requests field as-is with a fallback value? Same goes for the middleware. Using "as" in TypeScript gives me the ick, especially when the data comes from wherever.
```
// GLOBAL
const GH_ACTIONS = ["login", "signup", "link_account"];
...
const action = req.query.action;
if (typeof req.query.action !== "string" || !GH_ACTIONS.includes(req.query.action)) {
    return res.status(400).json({ ok: false, error: "Invalid action" });
}
```
??

frontend/src/user/login.page.tsx
EyzeCOLD, line 1: I think we have a mix of components with capital letter and lower letter initials? Should harmonize.

EyzeCOLD (changes requested):
Some esoteric requests for you to ponder, seems to function just fine.

backend/src/endpoints/oauth.ts
EvAvKein: Fixed - good catch!

backend/src/endpoints/oauth.ts
EvAvKein: Yup! Done

backend/src/endpoints/oauth.ts
EvAvKein: Added explicit guards for non-string token or username! After handling your comment on line 28, the timingSafeEquals throw is now handled too

backend/src/endpoints/oauth.ts
EvAvKein: Added the global allowlist and non-string guards to the endpoints. Had to still keep `as` for strong typing further down the program - the new guards will basically do runtime validations which'll make sure it fits into the type.

frontend/src/user/login.page.tsx
EvAvKein: We don't seem to have that on this file? But yes, we should make all components use PascalCase

EyzeCOLD (approved):
Yes.

PR #114, opened 2026-05-08
Syntax highlighting support for multiple languages

### LINKED ISSUE #101: CE: Editor multi-language syntax highlighting
Including default syntax by file extension, and a dropdown to override

EvAvKein:
Closes #101


frontend/src/codeEditor/Editor.tsx
```diff
@@ -220,9 +227,25 @@ export default function Editor({connection, myOwnerId, initialMembers, onRepickF
...
-					Change file
-				</Button>
+				<div className="flex items-center">
+					<select
+						className="m-1 px-1 border-2 border-surface bg-canvas text-foreground"
+						value={langSelected ?? ""}
+						onChange={(e) => setLangOverride(e.target.value || null)}
+					>
+						<option className="bg-canvas" value="">
+							File syntax
```
Omppu0: The dropdown shows 'File Syntax' always after loading the file even when the extension has been detected and syntax highlighting is active.
Also having the option off no highlighting with someting like 'Plain text' would be good. This should also be the default when extension is not detected

frontend/src/codeEditor/Editor.tsx
Omppu0, line 1: Basically all of the syntax highlighting has horrible contrast and does not look nice.

Omppu0 (changes requested):
So two thingies.
* The syntax highlighting does not look good or readable. 
* No no highlighting option and the dropdown isn't set properly on page load

Otherwise seems good

frontend/src/codeEditor/Editor.tsx
EvAvKein: Pushed commits to fix both!

frontend/src/codeEditor/Editor.tsx
EvAvKein: Was initially thinking to leave it to after/during Jukka's project-wide color pass, but since you mentioned it - I've set up the colors in this PR (which we can now more easily adjust later) ~

Omppu0 (approved):
Good enough for me 🙏

PR #115, opened 2026-05-08
Vim Mode

### LINKED ISSUE #108: CE: Vim mode

EvAvKein:
Toggleable in the Editor, settable as default in account settings.

The Vim toggle button in the editor isn't in its final spot, will be affected by PR #114 . Gave the Vim button a little glow when it's on ~

There's some Vim things that this plugin doesn't support even though it could, and I found out from the npm page that te Vim plugin is very configurable/extensible, but that's not for Cactus 😅

Closes #108


EyzeCOLD (approved):
Everything seems fine over here.

Is it possible to display the mode (INSERT, NORMAL, COMMAND, VISUAL etc) ? Would be really cool if.

PR #116, opened 2026-05-11
81 um enhancing um page

### LINKED ISSUE #56: UM - Frontend input validation
The input validation of forms in frontend could be improved.

### LINKED ISSUE #81: UM - Fixes requested in meeting
1. Have one submit button for all edits
2. Have user data be inputs by default
3. Request password for user deletion (and other changes)
4. Do not allow username/email/password to be same as before when changing
5. Show only hints for strong password and require only length
  - 'strong' password indicator? Optional

Sky11y:
**Items of issue #81**
1. Have one submit button for all edits
2. Have user data be inputs by default
3. Request password for user deletion (and other changes)
4. Do not allow username/email/password to be same as before when changing. Show only hints for strong password and require only length
5. 'strong' password indicator? Optional

**This PR handled the items as**
1. Finally decided to add different submit button for 'different' sections. Makes sense when scaling.
2. Only for username and email.
3. Add a Password component to all sections.
4. The checks done in frontend and backend. No hints. Required length of 8 characters.
5. Not implemented.

Closes #81 

Bonus: Add multiple input validations in frontend -> Closes #56


backend/src/endpoints/users.ts
```diff
@@ -43,27 +43,31 @@ function signupUser(app: Express) {
 function modifyUser(app: Express) {
 	app.patch("/api/user", requireAuth, async (req: Request, res: Response<ApiResponse<null>>) => {
 		timestampedLog(`REQUEST >>> ${req.method} ${req.url}`);
-		const {username, email, oldPassword, newPassword, vim_bindings} = req.body;
+		const {username, email, password, newPassword, newPassword2, vim_bindings} = req.body;
```
EyzeCOLD, line 46: Your changes to modifyUser seem to have broken the Vim mode toggle in the user management screen

backend/src/validation/schemas.ts
```diff
@@ -1,13 +1,7 @@
...
 import {z, type ZodType} from "zod";
 
-export const passwordSchema = z
-	.string()
-	.min(8, "Password must be at least 8 characters long")
-	.refine((password) => /[a-z]/.test(password), "Password must contain at least 1 small letter")
-	.refine((password) => /[A-Z]/.test(password), "Password must contain at least 1 capital letter")
-	.refine((password) => /[0-9]/.test(password), "password must contain at least 1 number")
-	.refine((password) => /[!@#$%^&*]/.test(password), "Password must contain at least 1 special character '!@#$%^&*'");
+export const passwordSchema = z.string().min(8, "Password must be at least 8 characters long");
```
EyzeCOLD: The consensus for the minimum length of a strong password is claimed in many places to be 14 all the way up to 16 characters.

frontend/src/user/userManagement.page.tsx
EyzeCOLD, line 1: This feels clunkier than before. It's odd that once the password confirmation prompt appears, you can't use the username or email fields anymore, even though there's no visual indication that they wouldn't be usable.

I think there's two options here:
A) Keep both fields editable until the user submits the changes with the password
B) Only allow one field to be edited at once. Have the password prompt appear under the one you're trying to change. Also keep the field editable until you confirm.

If you want to have the field "lock" when prompting for the password, make the field and the Button `disabled=true`. I'm not sure if the Input supports this yet, but it should be easy to implement (like it was done on the Button).

Also I don't think the password confirmation should say "Accept" but rather "Confirm" or "Verify" or something like that. The could also just be a Button next to Confirm.

EyzeCOLD (changes requested):
There was a regression bug and some small deets.

backend/src/endpoints/users.ts
Sky11y: All the other changes require password. So the password check before touching database broke it. Added a condition to the backend password if statement, since I don't think we need to require password for changing to vim keybindings.

backend/src/validation/schemas.ts
Sky11y: I changed it to 14.

frontend/src/user/userManagement.page.tsx
Sky11y: I agree. It is very clunky.

I removed the submit button => the password field activates when there's text in either of the fields. Also the fields can be modified even when the password confirmation is visible.
Maybe not as clunky anymore.

Sky11y: Changed the password section to behave same as username/email section.

EyzeCOLD (approved):
Yes.

PR #119, opened 2026-05-12
Frontend: accessibility and new login

### LINKED ISSUE #111: Increase text brightness for a18y contrast

EyzeCOLD:
* Improved accessibility
    * More contrast
    * Labels for inputs
* Component for TopBarLinks
* Styled the login page
* Added searchParams for /filebrowser

Closes #111

### DIFF
--- a/frontend/src/codeEditor/editorConfigs.ts
+++ b/frontend/src/codeEditor/editorConfigs.ts
@@ -237,11 +237,9 @@ export function getEditorExtensions({
 			},
 
 			// Styles the cursor's current line
-			"& .cm-activeLine, &.cm-merge-b .cm-activeLine": {
-				backgroundColor: "color-mix(in srgb, var(--color-accent) 7.5%, transparent)",
-			},
+			"& .cm-activeLine, &.cm-merge-b .cm-activeLine": {backgroundColor: "color-mix(in srgb, #cceeff44, transparent)"},
 			"& .cm-activeLineGutter, &.cm-merge-b .cm-activeLineGutter": {
-				backgroundColor: "color-mix(in srgb, var(--color-accent) 7.5%, transparent)",
+				backgroundColor: "color-mix(in srgb, #cceeff44, transparent)",
 			},
 
 			// Styles the button to expand section folds
--- a/frontend/src/components/TopBarLink.tsx
+++ b/frontend/src/components/TopBarLink.tsx
@@ -1,0 +1,25 @@
+import React from "react";
+import {NavLink, type NavLinkProps} from "react-router";
+
+type StyledNavLinkProps = Omit<NavLinkProps, "className"> & {
+	children: React.ReactNode;
+	className?: string;
+};
+
+const TopBarLink = React.forwardRef<HTMLAnchorElement, StyledNavLinkProps>(({children, className, ...props}, ref) => {
+	return (
+		<NavLink
+			ref={ref}
+			className={({isActive}) =>
+				`text-foreground-light px-3 py-2 hover:text-accent ${isActive ? "underline underline-offset-4 font-bold" : ""} ${className ?? ""}`
+			}
+			{...props}
+		>
+			{children}
+		</NavLink>
+	);
+});
+
+TopBarLink.displayName = "TopBarLink";
+
+export default TopBarLink;
--- a/frontend/src/components/TopBarLink.tsx
+++ b/frontend/src/components/TopBarLink.tsx
@@ -1,12 +1,7 @@
 import React from "react";
 import {NavLink, type NavLinkProps} from "react-router";
 
-type StyledNavLinkProps = Omit<NavLinkProps, "className"> & {
-	children: React.ReactNode;
-	className?: string;
-};
-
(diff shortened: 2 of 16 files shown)


EvAvKein:
The editor page's collab buttons and solo mode text still use the original (non-light) color, which is below the accessibility contrast threshold. Need to be switched to the new lighter color?

frontend/src/layout/Toast.tsx
EvAvKein, line 1: We shouldn't add these roles, they're actually worse for screenreaders for our use-case.

Per MDN's docs, the [`alert` role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/alert_role) and [`status` role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/status_role) are equivalent to setting `aria-live` (which we already do) and setting [`aria-atomic="true"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-atomic), and the latter makes the screenreader reread _the entire content of the element_. We don't want the screenreader to reread to the user all toasts every time we add just one.

frontend/src/user/login.page.tsx
```diff
@@ -50,38 +50,39 @@ export default function LoginPage() {
...
-						placeholder="password"
-						type="password"
-						value={loginPassword}
-						onChange={(e) => setLoginPassword(e.target.value)}
-					/>
-				</div>
-				<div>
-					<Button type="submit">Log In</Button>
-				</div>
+		<div className="flex flex-col items-center justify-center min-h-screen gap-2">
```
EvAvKein: The min-h-screen makes it so the page always has a scrollbar and empty scroll space. Could we maybe reduce the header size from the min height so it won't have that excess space at the bottom? That would also let the centered content be less off-center towards the bottom

frontend/src/components/TopBarLink.tsx
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import {NavLink, type NavLinkProps} from "react-router";
+
+type StyledNavLinkProps = Omit<NavLinkProps, "className"> & {
+	children: React.ReactNode;
+	className?: string;
+};
```
EvAvKein: Why is `className` being omitted and then readded via `&`?

EvAvKein (changes requested):

frontend/src/layout/Toast.tsx
EyzeCOLD: Alright, these were leftovers from fiddling with the Firefox's false positive. I removed the roles.

frontend/src/user/login.page.tsx
EyzeCOLD: I just made a fixed top padding so that it doesn't drift that low from the top bar. Also no extra emptiness.

frontend/src/components/TopBarLink.tsx
EyzeCOLD: Fixed this and other weird redundancies I found in other components.

frontend/src/codeEditor/editorConfigs.ts
```diff
@@ -237,11 +237,9 @@ export function getEditorExtensions({
 			},
 
 			// Styles the cursor's current line
-			"& .cm-activeLine, &.cm-merge-b .cm-activeLine": {
-				backgroundColor: "color-mix(in srgb, var(--color-accent) 7.5%, transparent)",
-			},
+			"& .cm-activeLine, &.cm-merge-b .cm-activeLine": {backgroundColor: "color-mix(in srgb, #cceeff44, transparent)"},
 			"& .cm-activeLineGutter, &.cm-merge-b .cm-activeLineGutter": {
-				backgroundColor: "color-mix(in srgb, var(--color-accent) 7.5%, transparent)",
+				backgroundColor: "color-mix(in srgb, #cceeff44, transparent)",
```
EvAvKein, line 242: Do we really want to replace the variable-based colors with hard-coded values? The variables were applying our project's theme

EvAvKein:
Also just realized that the tab instructions text for the editor don't pass color contrast, seems in-scope for this PR

frontend/src/user/login.page.tsx
EvAvKein: This kind of solution is generally not great for mobile, but we're planning to do a design pass anyway so I'll let this through

frontend/src/codeEditor/editorConfigs.ts
EvAvKein: Discussed in person - colors will be adjusted later, not relevant for this PR

EvAvKein (commented):
As discussed in person - the above comments aren't relevant right now, accessibility finishes continue on this branch

EvAvKein (approved):
I thought we decided in-person to label the selection element just "Syntax" instead of "Syntax Highlighting", and a lot of the added styles aren't doing that they they're attempting to do, but yeah those are not worth holding this PR back. Approving it, and we'll handle these cosmetic changes later as planned

PR #122, opened 2026-05-12
Frontend routing polish

### LINKED ISSUE #109: Frontend: 404 page
Page with a link, to a place

### LINKED ISSUE #110: Routing home pages updates
- Remove dashboard page, land logged-in users on filebrowser
- Add proper landing content (e.g. title splash screen, introduction) to `/` route

EvAvKein:
Closes #110 & #109 

Added some basic home page design, we can work on something less generic when we do the design pass


frontend/src/notFound.page.tsx
```diff
@@ -0,0 +1,24 @@
...
+			<p className="font-mono text-foreground/50 text-sm tracking-[0.35em] uppercase">404</p>
+			<h1 className="font-mono font-bold text-accent text-4xl">Page not found</h1>
+			{currentUser ? (
+				<Button onClick={() => navigate("/filebrowser")} className="mt-2 px-5 py-1.5 text-sm">
+					Go to files
+				</Button>
+			) : (
+				<Button onClick={() => navigate("/")} className="mt-2 px-5 py-1.5 text-sm">
+					Back to home
+				</Button>
```
EyzeCOLD: I think 404 should always only give a "Home" button, go to files feels a bit off. Also "Back TO Home" is a bit clunky english, even if it probably works here since home is jargon for home page and not literally home. So "Home" "Back Home" "Go Home" "Go Back Home" are your options basically.

frontend/src/layout/TopBars.tsx
EyzeCOLD, line 1: Not sure what file this pertains to, but the Top Bar is a bit borked when on the homepage. You only have the buttons "Home" and "Files" available. You should have Account and Log Out also available. Or more precisely, the top bar should give you all the same options no matter what page you're on.

We could also replace the Home button with the DiffEd logo at the topleft. I'm not sure how accessible that would be, but it'd be nice to have that floating there.

EyzeCOLD (changes requested):

frontend/src/notFound.page.tsx
EvAvKein: I'll shorten the button text, makes sense.

About the "files" option: Sending a user to the product landing page (the Home) feels a bit off to me - a logged-in user more likely wants to just use the product, and sending them to Home makes it so they need to click more buttons to get back to using the product.

How about having the Files button not as a replacement to Home, but as an addition?

frontend/src/layout/TopBars.tsx
EvAvKein: If we want to have the same buttons on all pages, we would have to combine the public and user topBars - right now the public topBar (which intentionally doesn't have the Account etc) is the one showing on all public pages.

I actually like the separation we currently have: Pages like Home are landing pages for non-users, and they have no user-specific functionality - personally it'd be a bit jarring to have a logout button on a public page, a page which is not specific to me (the user) in any way.

...I'd defer to a designer if we had one 🥴



Agree about changing Home to DiffEd ~

frontend/src/notFound.page.tsx
EyzeCOLD: There's three pages you can go: home, files and user management. I think we either go with Home or then just let people navigate using the top bar. Doesn't make sense to duplicate the top bar on the 404 page I think?

frontend/src/layout/TopBars.tsx
EyzeCOLD: Why can't we have it so that the top bar doesn't depend on which page you are on, but just on your login status? That makes the most sense to me.

It makes sense that when you're not logged in, we don't show you links that take you to unauthorized places.

However when you are logged in, it doesn't make sense to not give the user the same options for navigation on all pages. Even if the home page is quite useless for logged in users, we shouldn't punish them for going there by gimping their navigation options.

EvAvKein: We _can_ do what you're suggesting, I'm just of the opinion that it makes less UX sense.

This PR is just for doing the minimal routing changes we talked about in the last meeting, we didn't mention anything related to combining the separate bars. We've had the separation for the majority of the time we've had any topBar, I suppose it's just a bit more noticeable because there's fewer nav items. Should we decide the topic is out of scope for this PR, and talk about a potential topBar redesign/restructure as a team in person?

frontend/src/notFound.page.tsx
EvAvKein: If we want to have Home in the top bar (as per the other review comment), even having the Home button in the 404 would duplicate the top bar.

If we want to have nav buttons on the 404 page at all, to me it makes sense to give people two primary options - the Home page, and then if they're logging in the Files page (which, in its function, is the home page for a logged-in user).

EyzeCOLD: What I was going for, is that just having one button duplicates the top bar less. But also I don't see a need to have any additional buttons, since the top bar provides navigation already.

frontend/src/layout/TopBars.tsx
EyzeCOLD: Yeah I didn't notice it before or I just assumed it's something to do with the extra navigation we had before. But yeah we can have it be a separate PR too.

frontend/src/notFound.page.tsx
EvAvKein: Removed the buttons from the 404 page. PR should be ready for merge now?

EyzeCOLD (approved):

PR #131, opened 2026-05-13
104 public api

### LINKED ISSUE #104: Backend: Public API
- Public API key
  - Generate API key button
  - Regenerate button
  - Make `requireAuth` middleware accept that as a form of auth

Sky11y:
# Major changes
## backend
db
- add apikey to the users table

types
- create new type that expands the Express Request with userId

endpoints
- getUserApiKey (GET: "/api/user/api")
- updateUserApiKey (PATCH: "/api/user/api")

queries
- add getUserByApiKey
- add getApiKeyById
- add updateApiKey

Middleware
- add requireAuthOrApikey (used on endpoints where user can access information without logging in)

## frontend
user account page
- add generate new Api key button
- add copy api key button
- add delete api key button

# Minor Changes
- Update getUserById
- Some other cleanup

Closes #104


backend/src/endpoints/users.ts
```diff
@@ -22,9 +22,9 @@ function signupUser(app: Express) {
 				type: argon2.argon2id,
 			});
 
-			/* createUser returns id. Let's keep this call here if it is ever needed */
-			//const id = await createUser({username, email} as User, hash);
-			await userQueryService.createUser({username, email} as User, hash);
+			const id = await userQueryService.createUser(username, email, hash);
+			const user = await userQueryService.getUserById(id);
+			if (!user) throw new Error("User not found after creation");
```
EvAvKein: Is there a reason for querying the database to confirm the user was created? Is it actually possible for `createUser` to return an id without the user having been created?

backend/src/endpoints/users.ts
```diff
@@ -195,4 +195,48 @@ function getUser(app: Express) {
 	});
 }
 
-export default {signupUser, deleteUser, getUser, modifyUser};
+function getUserApiKey(app: Express) {
+	app.get("/api/user/api", requireAuthOrApiKey, async (req: AuthRequest, res: Response<ApiResponse<string>>) => {
+		timestampedLog(`REQUEST >>> ${req.method} ${req.url}`);
+
+		const id = req.session.userId! ?? req.userId;
```
EvAvKein: Why have the assertion that userId exists (the exclamation mark after it) if there's the nullish nullish coalescing (double question mark) after it?

backend/src/middleware.ts
```diff
@@ -8,3 +9,19 @@ export function requireAuth(req: Request, res: Response<ApiResponse<null>>, next
 		res.status(401).json({ok: false, error: "Unauthorized"});
 	}
 }
+
+export async function requireAuthOrApiKey(req: AuthRequest, res: Response<ApiResponse<null>>, next: NextFunction) {
+	if (req.session.userId) {
+		return next();
+	}
+
+	const apiKey = req.headers["x-api-key"] as string | undefined;
```
EvAvKein, line 22: The value can also be a string array, we should handle that with runtime logic instead of rejecting that option on the type level

backend/src/queries/users.ts
```diff
@@ -99,9 +121,9 @@ async function linkGithubId(userId: number, githubId: string): Promise<boolean>
 	return result.rowCount! > 0;
 }
 
-async function createUser(user: Omit<User, "id">, hash: string | undefined): Promise<number> {
+async function createUser(username: string, email: string, hash: string): Promise<number> {
 	const query = "INSERT INTO users (username, email, hashed_password) VALUES ($1, $2, $3) RETURNING id";
-	const values = [user.username, user.email, hash || null];
+	const values = [username, email, hash || null];
```
EvAvKein: If we're removing the possibility for the hash to be undefined in the parameter, should we also remove the `|| null` here?

frontend/src/user/userManagement.page.tsx
```diff
@@ -406,11 +406,110 @@ function Confirm({onConfirm, onCancel}: PasswordConfirmProps) {
...
+		try {
+			setIsLoading(true);
+
+			const hash = window.crypto.randomUUID();
+			const response: ApiResponse<string> = await apiFetch("/api/user/api", {
+				method: "PATCH",
+				headers: {"Content-Type": "application/json"},
+				credentials: "include",
+				body: JSON.stringify({hash}),
+			});
```
EvAvKein, line 422: API keys are usually UUIDs generated by the server, not strings set by the user the same way as a password.

It's a very big flaw that a user can set their own api key through a patch request to be simpler than it ought to be: If a user sets their API key to something overly simple, any other user who tries to query the API with that simple key gets access to through that user without even needing to know their username.

backend/src/endpoints/users.ts
```diff
@@ -195,4 +195,48 @@ function getUser(app: Express) {
 	});
 }
 
-export default {signupUser, deleteUser, getUser, modifyUser};
+function getUserApiKey(app: Express) {
+	app.get("/api/user/api", requireAuthOrApiKey, async (req: AuthRequest, res: Response<ApiResponse<string>>) => {
```
EvAvKein: `/api/user/api` is a bit of an awkward name - seems like the resource is the user's own API. Could we rename to something more descriptive, e.g. `/api/user/api-key`?

backend/src/middleware.ts
```diff
@@ -8,3 +9,19 @@ export function requireAuth(req: Request, res: Response<ApiResponse<null>>, next
 		res.status(401).json({ok: false, error: "Unauthorized"});
 	}
 }
+
+export async function requireAuthOrApiKey(req: AuthRequest, res: Response<ApiResponse<null>>, next: NextFunction) {
```
EvAvKein, line 17: The function is currently only used in `getUserApiKey`, but this should (would?) replace `requireAuth` at all endpoints.

With this prospect in mind I'm wondering if, instead of extending the Request type on all endpoints, we should see if there's a way to load the existing `req.session` with the `userId`? If direct assignment is too scary because it might interfere with our session-related dependencies, maybe there's a way to invoke the session dependencies to load user's session after we authenticate them by API key?

EvAvKein (changes requested):
Aside from the comments, there's apparently some backend compilation to fix - it found 2 errors and failed to compile for me 😅

backend/src/middleware.ts
Sky11y: This would only replace the requireAuth in endpoints where we allow user to touch the data either with api key or when logged in (active session). Those that allow only with active session would still use requireAuth as before -> no changes needed.

There might not be existing req.session for the user, who is using the api through curl or other app or whatever? Imo, involving session management would just make the api user experience way worse. Now user just needs to send the 'x-api-key' with the request(s) to the endpoints it is allowed to send them.

backend/src/endpoints/users.ts
Sky11y: yeah makes sense => changed as proposed.

frontend/src/user/userManagement.page.tsx
Sky11y: Yeah that was my initial thought, but for some reason I decided to create the hash at frontend without realizing that user could indeed send it's own hash, and we don't want that => hash is now created in backend.

backend/src/endpoints/users.ts
Sky11y: Not really. I guess it was just to double check if user creation was successful. I removed that.

backend/src/endpoints/users.ts
Sky11y: This was typescript thing, that's why it is everywhere. I changed the exclamation to be after req.userId, which makes more sense in this case.

backend/src/queries/users.ts
Sky11y: Wait.. I don't remember removing that :D It has to be possible to be undefined, since the user can create profile with OAuth as well...no wait, the OAuth user has different creation path => I removed the null.

Sky11y:
I also got the 2 errors after fclean. Note to self: always test with clean canvas...
The errors were a bit annoying so that I needed to change the logic, and now the AuthRequest type is directly given in middleware.ts, since that was the cleanest way I could figure out how to solve the errors.

backend/src/middleware.ts
Sky11y: Should never be an array of strings. If user send multiple strings server automatically concatenates to comma separated string, and the request fails (invalid api key) => no changes.

EvAvKein: Ah, right. I got mixed up with query parameters potentially being arrays.

backend/src/middleware.ts
EvAvKein: Fair point! Makes sense for it to stay as-is

EvAvKein (approved):
All issues fixed, good stuff!

PR #132, opened 2026-05-15
Frontend: FileBrowser & CodeEditor styling

EyzeCOLD:
FileBrowser and CodeEditor styling
* Code Editor
    * Syntax highlighting and diff coloring has been tweaked
  to mostly follow GitHub's look. However, the syntax parsing is
  fundamentally different, so some this is probably the best we can do without language specific rulesets
     * The filename change now resets the field `onBlur`, with a small exception to the Rename button with a `onMouseDown` trick
     * Added invisible borders to the Code Editors buttons so that they have the same height in and out of Vim Mode
* FileBrowser
    * Brought to the center of the screen and given a slight facelift. FilePicker and FileBrowser.page should now have the same look, except that FilePicker has a prominent header.
    * Added a danger variant to the Button for all sorts of deletion purposes
    * Added truncation to long filenames.


frontend/src/codeEditor/editorConfigs.ts
EvAvKein, line 1: A few problems I found with the theme changes:

- Changed lines have the red (self) background even when inside another user's changes. Probably need to add a `:not` selector somewhere to exclude lines inside approvable chunks (`reject` chunks, because of our proprietary inversion).
Image

- Highlighting doesn't work inside chunks that the other person doesn't have yet.
Image

- Vim mode still has very bright info bars (which I'm guessing is not a design choice, they're a bit bright and text doesn't have enough contrast).
- Whichever form of highlighting this, the contrast is not great... I'd hope something to be done about it in a way that would work across our text/background colors, but I'm not optimistic about that 😅
Image

frontend/src/codeEditor/Editor.tsx
```diff
@@ -275,21 +275,23 @@ export default function Editor({connection, myOwnerId, initialMembers, onRepickF
...
+						<Input
+							type="text"
+							value={newFileName}
+							onChange={(e) => setNewFileName(e.target.value)}
+							onBlur={() => setNewFileName(fileName)}
+						/>
 					</label>
-					<Button disabled={newFileName === fileName ? true : undefined} type="submit">
+					<Button
+						onMouseDown={(e) => e.preventDefault()}
```
EvAvKein: What is this `onMouseDown` `preventDefault`?

Does it do anything meaningful? If it does, it seems pretty generic - should we move that to inside the Button component?

frontend/src/codeEditor/FilePicker.tsx
```diff
@@ -12,8 +12,10 @@ function FilePicker({connection}: FilePickerProps): JSX.Element {
 	}
 
 	return (
-		<div className="p-4 flex flex-col gap-2">
-			<h2 className="text-lg">Pick a file to bring into this session</h2>
+		<div className="p-4 flex flex-col items-center justify-center gap-2">
+			<h2 className="text-xl w-max px-12 outline outline-offset-2 rounded-sm text-foreground-light bg-surface-dark">
```
EvAvKein, line 24: This feels a bit overstyled/loud. Do we want to try to design it more minimalistically in this PR, or leave that for the design pass?

frontend/src/components/Button.tsx
```diff
@@ -1,18 +1,22 @@
...
 
-type ButtonProps = React.ButtonHTMLAttributes<HTMLButtonElement>;
+type ButtonProps = React.ButtonHTMLAttributes<HTMLButtonElement> & {
+	danger?: boolean;
+};
 
-const disabledStyle = "bg-gray-700 text-stone-300";
+const disabledStyle = "bg-surface-dark text-foreground";
 const enabledStyle = "bg-surface text-foreground-light hover:text-accent cursor-pointer";
+const dangerStyle = "bg-canvas border border-surface-dark text-error hover:text-error-accent cursor-pointer";
```
EvAvKein: Apparently `text-error` on `bg-canvas` only has a 2.74 contrast ratio

frontend/src/fileBrowser/FileBrowser.tsx
```diff
@@ -24,12 +24,14 @@ function FileBrowser({onFileSelect}: fileBrowserProps) {
...
 
 	return (
-		<div className="flex flex-col center-items justify-center">
+		<div className="flex flex-col center-items justify-center gap-4">
 			{totalFiles > 0 ? (
-				<label>
-					Filter:
-					<Input type="text" id="fileListFilter" value={filter} onChange={(event) => setFilter(event.target.value)} />
-				</label>
+				<div className="flex center-items justify-center">
```
EvAvKein: Why not style the `label` directly instead of adding this `div`?

frontend/src/fileBrowser/FileList.tsx
```diff
@@ -55,23 +57,23 @@ function FileList({onFileSelect, fileList, refreshFileList, onSortToggle, descen
...
+				<td className="text-center w-fit">
 					<Button onClick={() => handleDownload(file)}>{" 🡻 "}</Button>
 				</td>
-				<td className="text-center w-24">
-					<Button onClick={() => handleDelete(file.id)} className="font-bold">
+				<td className="text-center w-fit">
+					<Button onClick={() => handleDelete(file.id)} className="font-bold" danger={true}>
 						{" X "}
 					</Button>
 				</td>
```
EvAvKein: With the `download` and `delete` table headers removed, does it still make sense to keep the buttons as separate table data? Also, no idea how this mismatch affects screenreaders.

Could reduce this to a single `<td>` per item. After that, I'd wonder if it'd make sense to just replace the table with an ordered list element 😅

frontend/src/fileBrowser/NewFile.tsx
```diff
@@ -37,23 +37,25 @@ function NewFile({onFileCreate}: NewFileProps): JSX.Element {
...
-				<Input
-					id="fileNameInput"
-					value={newFilename}
-					placeholder="filename"
-					onChange={(event) => setNewFilename(event.target.value)}
-				/>
-			</label>
-			<Button type="submit">Create</Button>
-		</form>
+		<div className="flex flex-col justify-center items-center">
```
EvAvKein, line 48: Same question as in the Button comment, but this time about `form` - instead of creating this parent element, why not style the child directly?

frontend/src/fileBrowser/PageSelector.tsx
EvAvKein, line 1: So now the "previous" and "next" buttons each have their own div and unordered list of a single item?
That's seems a bit unwieldy? And disorienting for screenreaders?

Could the structure maybe remain the same as the before, but give the button list items styling to achieve this same design?

frontend/src/layout/TopBars.tsx
```diff
@@ -66,7 +67,7 @@ export function UserTopBar() {
 				<TopBarLink to="/filebrowser">Files</TopBarLink>
 			</div>
 			<div className="flex items-center gap-2">
-				<TopBarLink to="/account">Account</TopBarLink>
+				<TopBarLink to="/account">{currentUser?.username}</TopBarLink>
```
EvAvKein: Doing optional chaining for the current user implies that it might be undefined, which would result in that being the username. Should we have a more user-friendly fallback?

EvAvKein (changes requested):
Besides the odd thing here and there, I really like the new design! Also, best of luck with the dive further into CodeMirror themeing 🫡

frontend/src/codeEditor/editorConfigs.ts
EyzeCOLD: I think I got all of it... Phew...

frontend/src/codeEditor/Editor.tsx
EyzeCOLD: I should probably document this, since it's not obvious what's going on. But the idea is that the rename input field has an `onBlur` that resets the name to what it was before you started fiddling with it. Now mouseDown is the event that triggers `onBlur` so without this prevention, you couldn't rename the file with the button, but only with enter.

EyzeCOLD: Added a comment

frontend/src/codeEditor/FilePicker.tsx
EyzeCOLD: I tried to make it very eye-catching. I think it's best you give it a shot later than me trying to guess.

frontend/src/components/Button.tsx
EyzeCOLD: pulled out red-500

frontend/src/fileBrowser/FileBrowser.tsx
EyzeCOLD: I tried, but it couldn't get it to look the same like that.

frontend/src/codeEditor/Editor.tsx
EvAvKein: With this implementation, it's impossible to rename the file via keyboard navigation. When pressing `tab` to move focus from the input to the button, the `onBlur` resets the edited name and the Rename button is disabled.

frontend/src/fileBrowser/FileList.tsx
EyzeCOLD: If there's no immediate accessibility problem with this, I'd rather not try to recreate the styling with different elements. :D

frontend/src/fileBrowser/NewFile.tsx
EyzeCOLD: I'm sure it can be done like that as well, but this is how I got it looking like I wanted to. The same styles don't seem to yield the same results in a different element.

frontend/src/fileBrowser/PageSelector.tsx
EyzeCOLD: Yeah the broken list was clearly a leftover from when it used to be just one. I cleaned it up so now there's just prev, next and a ul in between.

The pages are probably the thing that we want in the list, not the prev and next? They are kinda a thing of their own, right?

frontend/src/fileBrowser/FileBrowser.tsx
EvAvKein: I just tested, and apparently this looks exactly the same:

```
			{totalFiles > 0 ? (
				<label className="text-center">
					Filter
					<Input type="text" id="fileListFilter" value={filter} onChange={(event) => setFilter(event.target.value)} />
				</label>
			) : null}
```

frontend/src/layout/TopBars.tsx
EyzeCOLD: It should now show Account if things go wrong.

frontend/src/fileBrowser/FileList.tsx
EyzeCOLD: Okay it wasn't that hard, now it's a list.

frontend/src/fileBrowser/PageSelector.tsx
EvAvKein: Yeah, makes sense that prev and next are their own thing

frontend/src/index.css
```diff
@@ -37,9 +38,9 @@
 		background-color: var(--color-canvas);
 	}
 
-	@media (prefers-color-scheme: light) {
+	@media (prefers-color-scheme: dark) {
 		html {
-			color-scheme: light;
+			color-scheme: dark;
 		}
 	}
```
EvAvKein, line 45: Not important for this PR, but maybe we should just decide for the user that they're using a dark color scheme for this website (e.g. no media query, just set `color-scheme: dark`). We don't have a light-mode color palette and I don't think we're interested in supporting one in the forseeable future

frontend/src/codeEditor/editorConfigs.ts
EvAvKein: Unless the fixes aren't yet pushed - the highlighting in red blocks problem is still there, and same with the Vim mode info bars. The other two are fixed though!

EvAvKein (changes requested):

frontend/src/codeEditor/editorConfigs.ts
EyzeCOLD: Now both Vim bars should have `color-foreground-light` on `color-surface`.

Atleast on examples that I'm trying to the highlighting works for the red blocks as well. A few notes however: 
* if there's a diffed WORD then (the brighter red/green background) I'm overwriting the syntax highlight color for the word with white (like github does).
* some random words turn red in the diff, because the syntax highlighter treats those as some sort of syntax. I don't think there's a viable way to fix that.

frontend/src/codeEditor/Editor.tsx
EyzeCOLD: I created a component `ResettingForm` to encapsulate this behavior. So the answer was putting the onBlur on to the form and using `e.relatedTarget`. I should've gotten alarm bells for that hack I had before, seeing as this is quite a standard thing.

frontend/src/fileBrowser/FileBrowser.tsx
EyzeCOLD: Done. Thanks!

EyzeCOLD:
Changes!

frontend/src/codeEditor/editorConfigs.ts
EvAvKein: I'm not sure how to reproduce either of these. I'll bring it them next time we're meeting.

Created #139 for some other issue I found, I'm not sure how much it overlaps with what you're noting.

EvAvKein:
Created issues #137, #138, and #139 based on this PR, because it was starting to balloon out of scope. I'll try to PR fixes for those.

Reminder to self (and to you if you'll read this) to talk about the highlighting notes you mentioned - https://github.com/EyzeCOLD/DiffEd/pull/132#discussion_r3258689609

EvAvKein (approved):

PR #136, opened 2026-05-17
98 user avatars

### LINKED ISSUE #98: UM: User avatars
- Image POST (frontend & backend)
- Image GET  (frontend & backend)
- Image DELETE (frontend & backend)
- Image validation (frontend & backend)
- Image preview for upload
- Make user avatar visible in top bar and editor member list

### LINKED ISSUE #141: UM: Avatar upload tip
The avatar upload should be explicit about the type and size restrictions on the page.

Sky11y:
# Add avatars to user account page

### database
- add avatar_filename to users
- add two new queries: getAvatarFilenameById, updateAvatar

### Endpoints
- New endpoint "/api/user/avatar" => file backend/src/endpoint/avatar.ts
- Add avatar removal to "api/user" DELETE endpoint

### frontend
- Add Avatar component on userManagement.page.tsx
- Create accept prompt component
  - Add it to api key deletion and avatar deletion

### types
- add has_avatar boolean to user type

## Other
- Separated api-key endpoints out of user endpoint => file backend/src/endpoints/apikey.ts
- Add DELETE endpoint to "api/user/api-key"
  - Fixed a bug related to that, api-key was persisted though user requested its' deletion.

Closes #98
Closes #141


backend/src/endpoints/apiKey.ts
EyzeCOLD, line 1: Filename should probably be capitalized `apiKey` like elsewhere.

Also was there a some sort of design idea behind having the getter be usable via API key, but not have the patch and delete work the same way?

backend/src/endpoints/apikey.ts
```diff
@@ -0,0 +1,73 @@
...
+		try {
+			const key = await userQueryService.getApiKeyById(id);
+			if (!key) {
+				return res.status(404).json({ok: false, error: "NO_API_KEY"});
+			}
+
+			res.status(200).json({ok: true, data: key});
+		} catch (error: unknown) {
+			if (isDbError(error)) {
+				timestampedLog(`ERROR <<< ${error.code}: ${error.detail}`);
```
EyzeCOLD: Could add the `DB ERROR` to the log, like elsewhere in the codebase for DB errors.

backend/src/endpoints/avatars.ts
```diff
@@ -0,0 +1,160 @@
...
+import path from "path";
+import fs from "fs";
+import type {Express, Request, Response, NextFunction} from "express";
+import {isDbError} from "#/src/utils.js";
+import type {ApiResponse} from "#shared/src/types.js";
+import {timestampedLog} from "#/src/logging.js";
+import {requireAuth} from "#/src/middleware.js";
+import userQueryService from "#/src/queries/users.js";
+
+const MAX_IMGSIZE = 1024 * 1024; // 1 MB
```
EyzeCOLD: This is a MiB, MebiByte. We're using actual MB in the files (just a simple million). Doesn't really matter. Just saying.

backend/src/endpoints/avatars.ts
```diff
@@ -0,0 +1,160 @@
...
+		next();
+	});
+}
+
+function updateUserAvatar(app: Express) {
+	app.patch("/api/user/avatar", requireAuth, multerUpload, async (req: Request, res: Response<ApiResponse<null>>) => {
+		timestampedLog(`REQUEST >>> ${req.method} ${req.url}`);
+
+		if (!req.file) {
+			return res.status(400).json({ok: false, error: "Invalid filetype or size"});
```
EyzeCOLD: How do we end up here? Aren't these errors already handled by `multerUpload`?

backend/src/endpoints/avatars.ts
```diff
@@ -0,0 +1,160 @@
...
+			const old_filename = await userQueryService.getAvatarFilenameById(id);
+			if ((await userQueryService.updateAvatar(filename, id)) === false) {
+				throw new Error(`Could not update user avatar with id: ${id}`);
+			}
+
+			if (old_filename && old_filename !== "empty.jpg") {
+				deleteAvatar(old_filename);
+			}
+			res.status(200).json({ok: true, data: null});
+		} catch (error: unknown) {
```
EyzeCOLD, line 91: If anything throws, isn't it possible that the file is already uploaded and just stays in storage forever?

backend/src/private/avatars/default.jpg
EyzeCOLD, line 1: `default` might be a better name?

frontend/src/user/userManagement.page.tsx
```diff
@@ -407,8 +407,27 @@ function Confirm({onConfirm, onCancel}: PasswordConfirmProps) {
 	);
 }
 
+type AcceptProps = {
+	onAccept: () => void;
+	onCancel: () => void;
+};
+
+function PromptAccept({onAccept, onCancel}: AcceptProps) {
```
EyzeCOLD: I think it would be neat to have this in `/components/`, since it's quite self-contained and reusable. Some of these files are getting very long.

frontend/src/user/userManagement.page.tsx
```diff
@@ -494,16 +513,113 @@ function ApiKey({hasApiKey}: {hasApiKey: boolean}) {
...
+							</div>
+						)}
 					</>
 				) : null}
 			</div>
 		</>
 	);
 }
 
+function Avatar({hasAvatar}: {hasAvatar: boolean}) {
```
EyzeCOLD, line 468: This too could just be in its own file.

frontend/src/user/userManagement.page.tsx
```diff
@@ -494,16 +513,113 @@ function ApiKey({hasApiKey}: {hasApiKey: boolean}) {
...
+	const [avatarURL, setAvatarURL] = useState("/api/user/avatar");
```
EyzeCOLD: Should we use the timestamp trick here as well? Now if the avatar is changed, the change won't be noticeable on component mount. Only on actual F5. So if you change your avatar, then go Files and back, you have a stale avatar.

I also read that there are "more professional" ways to do this, but I don't know if we care at this stage.

EyzeCOLD (changes requested):
Small things.

backend/src/endpoints/apiKey.ts
Sky11y: The idea was only to have an example in the codebase how to implement an endpoint where user can reach with api key. And also allow reviewer to test the requireAuthOrApiKey and AuthRequest without writing a line of code. I'm down to remove this from the endpoint. It doesn't serve the user too much.

backend/src/endpoints/avatars.ts
Sky11y: Good catch. Yeah we shouldn't end up here. This is a leftover from earlier implementation where middleware multerUpload was not implemented. Removed the if statement.

backend/src/endpoints/apikey.ts
Sky11y: Added DB ERROR to files apiKey.ts, avatars.ts, and users.ts

backend/src/endpoints/avatars.ts
Sky11y: True. Changed the comment.

backend/src/private/avatars/default.jpg
Sky11y: \+ I already had the constant value DEFAULT_AVATAR for it, so changed that to everywhere.

backend/src/endpoints/avatars.ts
Sky11y: True! Added deleteAvatar(filename) to the catch block, and also added try-catch block for the deleteAvatar since fs (and path) can throw as well.

frontend/src/user/userManagement.page.tsx
Sky11y: Add the trick here as well.

frontend/src/user/userManagement.page.tsx
Sky11y: Yeah, separated this and PasswordConfirm to their own files /components/

frontend/src/user/userManagement.page.tsx
Sky11y: I think it makes sense to keep all the directly related to this page in this file, and only separate components that can be reused on other pages, e.g. AcceptCancelDialog and PasswordConfirmDialog, that are now separated to their own components

EyzeCOLD: Fair enough.

EyzeCOLD (approved):

PR #143, opened 2026-05-19
Changed CodeMirror text selection to use system styling

### LINKED ISSUE #139: Make CodeMirror cursor-based highlighting more consistent
Right now when highlighting CodeMirror content while dragging the cursor across some text, it:
- Highlights the Accept-able (green chunk) text and the Accept button with the browser's default text highlighting.
- Highlight non-diffed text with our custom editor highlighting theme.
- Not highlight replaceable (red chunk) text at all.

Should probably make them consistent, using the CodeMirror theming.

EvAvKein:
Replaced our custom .cm-selectionBackground theming with a higher-specificity ::selection rule, that overrides CodeMirror's built-in transparent rule and sets selection to use the system selection colours (Highlight / HighlightText). This shows more uniformly between plain code, green (acceptable) chunks, and red (replaceable) chunks, without us having to fiddle with more themeing and syntax-highlighting edge-cases.

Closes #139


EyzeCOLD:
The new consistency for the selection is cool. Very readable.
image

We should probably do something similar for the search matches also, since atleast the current search match doesn't have proper contrast with atleast the red text. Could do a different color maybe for the matches, but have the same idea where the syntax highlighting for the text is overridden so that we don't have to worry about all combinations for contrast. There's a search in Vim Mode with `/` and also a CM builtin search you can see on the screenshot that's accessible via C-f while in the editor

Then there's the third case, the highlight selection match case, so if you paint a word with the cursor, it will highlight every other instance. I'd use the same trick here (override the syntax hl color) but maybe again use a different color, so it's clear what you are selecting and what are the matches.

EvAvKein:
> We should probably do something similar for the search matches also, since atleast the current search match doesn't have proper contrast with atleast the red text. Could do a different color maybe for the matches, but have the same idea where the syntax highlighting for the text is overridden so that we don't have to worry about all combinations for contrast. There's a search in Vim Mode with `/` and also a CM builtin search you can see on the screenshot that's accessible via C-f while in the editor
> 
> Then there's the third case, the highlight selection match case, so if you paint a word with the cursor, it will highlight every other instance. I'd use the same trick here (override the syntax hl color) but maybe again use a different color, so it's clear what you are selecting and what are the matches.

Done! Considering how much better the system styling looks compared to selecting our own colors, I opted to keep using system colors - used lower opacities to help distinguish between these cases.

EyzeCOLD (approved):
Phew..!

PR #147, opened 2026-05-21
Convert legal pages to markdown-first

EvAvKein:
Moving the legal docs to markdwon-first, to be in line with the API docs


Sky11y (approved):
LGTM

PR #148, opened 2026-05-21
Frontend: User Management styling + bonus

### LINKED ISSUE #126: BUG: NewFile with long filename
You get a database error if you try to create a new file with a very long filename.

Most likely replicable with all user fillable fields, like username, password, email...

EyzeCOLD:
* Styled the user management page
* Improved NewFile and Code Editor's file rename with validation
* Made DeleteButton.tsx to replace `window.confirm()` call in filelist
* Made multiple buttons be disabled when there's no good reason to click on them


shared/src/fileValidation.ts
```diff
@@ -1,22 +1,37 @@
...
+const MAX_FILENAME_LEN = 100;
+const TYPE_BLACK_LIST = ["image/", "video/", "audio/", "font/"];
+const NAME_CHAR_BLACKLIST = ["\0"];
+
+export function validateFileNameLen(fileName: string): string | null {
+	const trimmed = fileName.trim().length;
+	if (trimmed === 0) {
+		return "Filename can't be empty";
+	}
+	if (trimmed > MAX_FILENAME_LEN) {
```
EvAvKein: If you're validating against the trimmed version, a filename with a single character and way too many spaces can pass validation and cause the #126 error. Best to use the untrimmed length for this check

backend/src/validation/schemas.ts
```diff
@@ -23,7 +23,6 @@ export const SignupSchema = z.object({
 	password: passwordSchema,
 }) satisfies ZodType<SigningUser>;
 
-// id here for testing
 export const UserFileSchema = z.object({
 	id: z.uuidv4(),
 	name: z.string(),
```
EvAvKein: Could add the new `MAX_FILENAME_LEN` length limit as part of this schema?

frontend/src/components/ResettingForm.tsx
```diff
@@ -30,9 +34,10 @@ function ResettingForm({initialValue, onSubmit, inputLabel, buttonLabel}: resett
 				{inputLabel}
 				<Input type="text" value={draft} onChange={(e) => setDraft(e.target.value)} />
 			</label>
-			<Button type="submit" disabled={initialValue === draft ? true : undefined}>
+			<Button type="submit" disabled={buttonDisabled}>
 				{buttonLabel ?? "Submit"}
 			</Button>
+			{validationError ? <div className="text-sm text-error-accent">{validationError}</div> : null}
```
EvAvKein, line 40: Would be nice if the message was centered by the parent element, it's currently misaligned e.g. in file renaming

frontend/src/components/DeleteButton.tsx
```diff
@@ -0,0 +1,39 @@
+import {useState, useEffect} from "react";
+import Button from "./Button";
+import {IconTrash, IconCheck, IconX} from "@tabler/icons-react";
+
+type deleteButtonProps = {
+	onDelete: () => void;
+	fileName: string;
+};
+
+function DeleteButton({onDelete, fileName}: deleteButtonProps) {
```
EvAvKein: For accessiblity/screenreaders, it would be good if the element always returned a live-region (assertive), and the content of the live-region is what's depending on the `confirming`

frontend/src/components/DeleteButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+	const [confirming, setConfirming] = useState(false);
+
+	useEffect(() => {
+		if (!confirming) return;
+		const timer = setTimeout(() => setConfirming(false), 3000);
+		return () => clearTimeout(timer);
+	}, [confirming]);
```
EvAvKein: Is there any benefit for having this timeout? I get the concept of a reset to keep the UI clean, but I'd say giving users time and the plain option to dismiss the deletion on their own is better. There's some subjectivity to the UX of course, but not for the accessibility aspect (especially after adding the live-region).

frontend/src/fileBrowser/NewFile.tsx
```diff
@@ -52,16 +54,21 @@ function NewFile({onFileCreate, refreshFileList}: NewFileProps): JSX.Element {
 					openNewFile();
 				}}
 			>
+				{newFileName.length > 0 && !!fileNameError ? (
+					<div className="flex justify-center text-sm text-error">{fileNameError}</div>
+				) : null}
```
EvAvKein: ARIA live region for errors?

frontend/src/user/userManagement.page.tsx
```diff
@@ -343,16 +341,25 @@ function Delete() {
...
-				Delete Account
-			</Button>
-		</div>
-	) : (
-		<div>
-			<h2>You are deleting your account!</h2>
-			<PasswordConfirmDialog onConfirm={deleteAccount} onCancel={() => setPasswordConfirm(false)} />
+	return (
+		<div className="flex flex-col justify-center items-center outline-1 outline-error-accent rounded-sm p-2 m-2">
+			<Header className="text-error-accent">Danger Zone</Header>
```
EvAvKein: This header component is an `h1`, those are supposed to be used for the content of a page ([MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/Heading_Elements#avoid_using_multiple_h1_elements_on_one_page)). It's not good semantic HTML to use a header with this level to title a subsection

frontend/src/user/userManagement.page.tsx
```diff
@@ -343,16 +341,25 @@ function Delete() {
...
-			<Button onClick={() => setPasswordConfirm(true)} aria-label="Delete account">
-				Delete Account
-			</Button>
-		</div>
-	) : (
-		<div>
-			<h2>You are deleting your account!</h2>
-			<PasswordConfirmDialog onConfirm={deleteAccount} onCancel={() => setPasswordConfirm(false)} />
+	return (
+		<div className="flex flex-col justify-center items-center outline-1 outline-error-accent rounded-sm p-2 m-2">
```
EvAvKein, line 328: Same as the other comments - screenreaders won't announce changing to this if there's no screenreader live-region

frontend/src/user/userManagement.page.tsx
EvAvKein: Also not good that this `h1` component is used multiple times in this page file. See link ^

(Maybe a good solution to both would be to change it to `h2` and rename it to `Subheading`?)

frontend/src/user/userManagement.page.tsx
```diff
@@ -343,16 +341,25 @@ function Delete() {
...
+			<Header className="text-error-accent">Danger Zone</Header>
+			{!passwordConfirm ? (
```
EvAvKein: Don't wanna repeat myself too much, so I'll just add a small note "live-region" for the other places I spot the situation

frontend/src/user/userManagement.page.tsx
```diff
@@ -435,33 +442,26 @@ function ApiKey({hasApiKey}: {hasApiKey: boolean}) {
...
-					Create new Api key
+		<OutlineDiv>
+			<Header>API Keys</Header>
+			<Button onClick={createNewApiKey} disabled={isLoading} aria-label="Create new API key">
+				Create new API key
+			</Button>
+			<Button onClick={copyApiKey} disabled={isLoading || !hasApiKey} aria-label="Get current API key">
+				Copy current API key
+			</Button>
+			{showDeleteConfirm ? (
```
EvAvKein: live-region

frontend/src/user/userManagement.page.tsx
```diff
@@ -532,30 +532,30 @@ function Avatar({hasAvatar}: {hasAvatar: boolean}) {
...
+			<div className="w-32 h-32 rounded-full overflow-hidden m-2">
 				<img src={avatarURL} alt="avatar" className="w-full h-full object-cover" />
 			</div>
 			<input type="file" accept="image/*" onChange={handleUpload} ref={fileInputRef} style={{display: "none"}} />
 			<Button disabled={isLoading} onClick={() => fileInputRef.current?.click()}>
 				Upload avatar
 			</Button>
-			<span>max. 1 MB (.png .webp .jpg .jpeg)</span>
-
 			{hasAvatar && !acceptDelete ? (
```
EvAvKein: live-region

frontend/src/user/userManagement.page.tsx
```diff
@@ -583,29 +583,25 @@ export default function UserManagementPage() {
...
-				<UserSettings user={currentUser} onUpdate={(username, email) => updateUser({username, email})} />
-			</div>
-			<div>
-				<Password />
-			</div>
-			<div>
-				<GithubLink githubLinked={!!currentUser.github_linked} />
-			</div>
-			<div>
+		<div className="mx-auto w-fit">
```
EvAvKein: Now that this div's width is down to `w-fit`, the footer inside it gets shrunk. Can the footer be moved outside this div?

frontend/src/components/DeleteButton.tsx
EvAvKein: Since it's in the general "components" folder, can it be renamed to something more descriptive? It sounds like a general button for deletion functionalities, but it's just for files

frontend/src/components/AcceptCancelDialog.tsx
```diff
@@ -8,14 +8,16 @@ type AcceptCancelProps = {
...
 	return (
-		<div>
-			<span>{textToShow}</span>
-			<Button onClick={onAccept} aria-label="Accept changes">
-				Accept{" "}
-			</Button>
-			<Button onClick={onCancel} aria-label="Cancel changes">
-				Cancel{" "}
-			</Button>
+		<div className="flex flex-col">
```
EvAvKein: Does the flexbox do anything here?

frontend/src/components/Header.tsx
```diff
@@ -0,0 +1,12 @@
+import type {ReactNode} from "react";
+
+type headerProps = {
+	className?: string;
+	children?: ReactNode;
+};
+
+function Header({className, children}: headerProps) {
```
EvAvKein: Left some comments in the review userManagement page regarding this component

frontend/src/components/PasswordConfirmDialog.tsx
```diff
@@ -29,28 +29,19 @@ export default function PasswordConfirmDialog({onConfirm, onCancel}: PasswordCon
 	}
 
 	return (
-		<>
+		<div className="flex flex-col justify-center items-center">
+			<label>
+				Password
```
EvAvKein: This label being on the left of the input looks a bit weird in the account settings page, where the other labels are above their respective inputs

EvAvKein (dismissed):

backend/src/validation/schemas.ts
EyzeCOLD: Done.

The 100 actually comes from `files.sql`, I wonder if there would be any sensible way to have all of these use the same constant.

shared/src/fileValidation.ts
EyzeCOLD: Fixed

frontend/src/components/ResettingForm.tsx
EyzeCOLD: slapped a "items-center" on that flex

frontend/src/components/DeleteButton.tsx
EyzeCOLD: Done.

Even though I'm not sure if this matters, since when you press the button, the focus will automatically be on the "confirm" button so the reader will announce that regardless of the aria-live.

frontend/src/components/DeleteButton.tsx
EyzeCOLD: Yeah I actually hate it too. AI pitched it and it sounded like a reasonable idea for a second.

frontend/src/fileBrowser/NewFile.tsx
EyzeCOLD: I tried.

frontend/src/user/userManagement.page.tsx
EyzeCOLD: Renamed and downgraded to h2

backend/src/validation/schemas.ts
EvAvKein: We could add the constraints outside of the SQL by altering the tables through our `postgres.ts`, don't know how sensible that is

EvAvKein:
As discussed, I took of a few final things, and the rest you resolved in commits today! Good stuff, now we just need someone who hasn't committed in this PR to push the merge button :sweat_smile:

frontend/src/user/userManagement.page.tsx
Omppu0: Seems this comment was left dangling. I looked trough the code and seems like the live region required for this one is now inside the PasswordConfirmDialog.

Omppu0 (approved):
Trusting things are fine here and this has already been reviewed as discussed.

PR #149, opened 2026-05-22
Pre-evaluation polish

### LINKED ISSUE #107: Polish for eval
- Remove `console.error` in frontend (or convert to `log` :) )
- Remove unused `change_date` from users SQL
- Set max workspace members capacity

Bonus (not for eval):
- Polish wiki, make sure it's up to date

EvAvKein:
Opened up the GET workspace endpoints to non-members as well, for the API. Now you don't have to connect with a socket to be able to query this (but you still can't see people's files without being a member with a socket connection).

Closes #107


frontend/src/codeEditor/peerExtension.ts
```diff
@@ -30,8 +30,8 @@ export function peerExtension(startVersion: number, connection: CollabConnection
 			const version = getSyncedVersion(this.view.state);
 			try {
 				await pushUpdates(connection, version, updates);
-			} catch (error) {
-				console.error("Failed to push updates:", error);
+			} catch {
+				console.log("Failed to push updates. Try another!");
```
Sky11y, line 34: Who is this log for? I doubt basic user would look at dev tools?

Sky11y (approved):
Solid stuff.

frontend/src/codeEditor/peerExtension.ts
Sky11y: This was discussed personally. Let's keep the log here, so catch block has something in it.

PR #152, opened 2026-05-22
fixed the file get endpoints to only return the needed data

### LINKED ISSUE #151: Make api/files get request return only needed data

Omppu0:


EvAvKein (approved):
The type of the response value for this endpoint is out of sync, but that's not too high of a priority. Usually I'd ask for that to be fixed in the same PR, but considering our team's upcoming availability (lack thereof) and evaluation timetable, the syncing is a really low priority.

PR #156, opened 2026-05-26
Account-related fixes

EvAvKein:
- Add frontend password validation on signup and password-change forms using a shared `validatePassword` helper
- Add frontend avatar file-size check (1 MiB) before upload, matching the backend limit
- Extract `PASSWORD_MIN_LENGTH` and `MAX_AVATAR_SIZE` into a new `shared/src/userValidation.ts`
- Fix `UserSettings` so `onUpdate` properly uses and updates the user store
- Fix `PasswordConfirmDialog` to use the `disabled` prop on its final submit button
- Fix `NewFile.tsx` confirmation function throwing due to undefined (miscapitalized) variable


backend/src/endpoints/avatars.ts
```diff
@@ -4,11 +4,11 @@ import fs from "fs";
 import type {Express, Request, Response, NextFunction} from "express";
 import {isDbError} from "#/src/utils.js";
 import type {ApiResponse} from "#shared/src/types.js";
+import {MAX_AVATAR_SIZE} from "#shared/src/fileValidation.js";
```
Sky11y: should be userValidation.js, not fileValidation.js

shared/src/userValidation.ts
Sky11y, line 1: We discussed with Jukka to limit the max password length as well. Should be easy to implement here already. 128 chars should be enough.

backend/src/validation/schemas.ts
Sky11y, line 1: import the PASSWORD_MAX_LENGTH here and add it to the passwordSchema `Password can not be longer than ${PASSWORD_MAX_LENGTH} characters.`

frontend/src/user/userManagement.page.tsx
Sky11y: remove this line, since we don't want to make user clipboard dirty without user 'approval'.

frontend/src/components/PasswordConfirmDialog.tsx
Sky11y: We need to somehow reset the state when the password and change has been accepted. Should we require a return value from onConfirm functions?

if (onConfirm === true) {
  setPassword("");
  setIsConfirming(false);
}

We could also just add the lines there without if-statement, but it would be a bit on nuisance to the user to always write the password again, even with minor typo.

Sky11y (changes requested):
Nice job. I added some comments that are related to changed files, and that could be fixed with this PR.

backend/src/endpoints/avatars.ts
EyzeCOLD: Yes

shared/src/userValidation.ts
EyzeCOLD: Fixed

backend/src/validation/schemas.ts
EyzeCOLD: Yes.

frontend/src/user/userManagement.page.tsx
EyzeCOLD: yes

frontend/src/components/PasswordConfirmDialog.tsx
EyzeCOLD: Yes

backend/src/endpoints/avatars.ts
EvAvKein: Oops, had that fixed locally but forgot to push. Done!

Sky11y (changes requested):
Seems to be working but the Github checks fail.

Sky11y (approved):
LGTM and go MS!

PR #161, opened 2026-05-26
Update README with new dependencies and clarifications

Sky11y:
Added express-rate-limit and helmet to dependencies. Clarified user session management and added details about user avatar storage. Updated challenges faced by the team.


Sky11y:
Maybe we should not mention single functions people have implemented in the `contributions` section?

EvAvKein:
> Maybe we should not mention single functions people have implemented in the `contributions` section?

Noted! As discussed, I'll remove the function names from that section in my upcoming PR

EvAvKein (approved):

PR #169, opened 2026-05-27
Big-file handling: Backend checks simplification, editing/socket prevention

### LINKED ISSUE #158: Files being too large are caught by NGINX
NGINX responds with 413, which we want, but it doesn't respond with a ApiResponse. We could just add it to ApiFetch, that when it gets 413, it responsds with `{ok: false, error: "File too large"}`

If we let NGINX handle the big files, should we remove redundant checks from the backend?

EvAvKein:
For endpoints: Made the existing NGINX file-size check explicit and removed redundant server checks.

For collaborative editing: Added an edit hook to reject any edits that grow past the size limit, and also added a server-side check.

Closes #158


shared/src/fileValidation.ts
```diff
@@ -1,7 +1,7 @@
-const MAX_FILE_SIZE = 1000000; // 1meg
 const TYPE_BLACK_LIST = ["image/", "video/", "audio/", "font/"];
 const NAME_CHAR_BLACKLIST = ["\0"];
 export const MAX_FILENAME_LEN = 100;
+export const MAX_FILE_SIZE = 1048576; // 1 MiB. Keep this in sync with NGINX's client_max_body_size, which is the actual backend limiter
```
Sky11y: As discussed, let's change this to 1000000

backend/nginx/nginx.conf.template
```diff
@@ -7,6 +7,9 @@ http {
     server backend:3000;
   }
 
+  # Single source of truth for backend upload size - keep this in sync with frontend upload limits
```
Sky11y: It is not in sync with upload limits, this seems to be 1MB and frontend has 1MiB

frontend/src/utils.ts
```diff
@@ -15,6 +15,10 @@ export async function apiFetch<T>(input: RequestInfo, init?: RequestInit): Promi
 		return {ok: false, error: "Network error"};
 	}
 
+	if (res.status === 413) {
```
Sky11y, line 18: This could also be in the catch block below like the other res.status === 429 case

backend/src/server.ts
```diff
@@ -16,13 +16,14 @@ import OAuthEndpoints from "./endpoints/oauth.js";
...
 import "./passportConfig.js";
+import {MAX_FILE_SIZE} from "#shared/src/fileValidation.js";
 
 import {initCollabSocket} from "./endpoints/collabSocket.js";
 
 const app = express();
 app.set("trust proxy", 1); // Required so Express reads the Host header forwarded by Nginx, needed for passport-oauth2 relative callbackURL resolution
 const server = createServer(app);
-const sockets = new Server(server, {cors: {origin: "*"}});
+const sockets = new Server(server, {cors: {origin: "*"}, maxHttpBufferSize: MAX_FILE_SIZE * 2}); // Doubles MAX_FILE_SIZE so payloads near the file-size limit still fit after JSON envelope and string-escape overhead
```
Sky11y, line 26: Not sure how this works if nginx already rejects the size?

backend/src/endpoints/avatars.ts
Sky11y, line 1: I guess it wouldn't hurt to just change the MAX_AVATAR_SIZE to MAX_FILE_SIZE and leave the checks here?

backend/nginx/nginx.conf.template
Sky11y: As discussed this is ok. You can ignore the previous comment.

backend/src/endpoints/files.ts
```diff
@@ -77,7 +77,7 @@ function uploadFile(app: Express) {
 			const fileName = f.originalname;
 			const fileContent = f.buffer.toString("utf8");
 
-			const err: string | null = validateFile(f.mimetype, f.size, fileContent, fileName);
```
Sky11y, line 80: Why take this away?

Sky11y (changes requested):
gg. nginx should return ApiError json. I don't remember if i commented it separately but here you go.

Sky11y (changes requested):
Tested and proven nginx is MiB and counts the whole request, not just the body. Let's keep this `  client_max_body_size 1m;` as 1m and change MAX_FILE_SIZE = 1'000'000

backend/src/endpoints/avatars.ts
EvAvKein: Discussed - yup!

backend/src/endpoints/files.ts
EvAvKein: Discussed - readded

frontend/src/utils.ts
Sky11y: As discussed, this is okay.

backend/src/server.ts
Sky11y: Had a nice discussion about sizes, and resolved this.

Sky11y (approved):
LGTM

PR #171, opened 2026-05-27
NGINX config and rate limiting

### LINKED ISSUE #168: Cors policy issue and page redirect when connecting from non host computers
Possibly only affects the dev build but will see.

### LINKED ISSUE #170: BUG: Ratelimiting doesn't discern different IPS
Rate limiter in login endpoint was treating every connection as the same client, which would lock out users from logging in if over 5 logins were made in 15 minutes.

### LINKED ISSUE #172: Default avatar not working on prod
Might be an issue with the paths?

EyzeCOLD:
* Moved to nginx rate limiting. Separate limits for sockets and api.
* Took out the login rate limiter
* Made ssl_ stricter in prod, but also added support for TLSv1.2 (compatibility)
* Prod ports are now set in backend/.env HTTP_PORT, HTTPS_PORT

Closes #170 
Closes #168 
Closes #172


Sky11y:
Is there a reason why the HTTPS ports are 433 instead of 443?

backend/dockerfile.dev
```diff
@@ -12,7 +12,9 @@ RUN npm install
 # BUILD
 COPY /shared /shared/
 COPY /backend /backend/
+RUN mkdir -p /data/avatars
+COPY /backend/src/defaultAvatar.jpg /data/avatars/default.jpg
 
 # HOST
+CMD ["npm", "run", "buildDevAndServerDev"]
 EXPOSE 3000
```
Sky11y: Could remove since no use for EXPOSE

backend/dockerfile
```diff
@@ -26,7 +26,9 @@ RUN npm install --omit=dev
 
 COPY /backend/sql/ /backend/sql/
 COPY /backend/.env /backend/.env
+RUN mkdir -p /data/avatars
+COPY /backend/src/defaultAvatar.jpg /data/avatars/default.jpg
 
 # HOST
+CMD ["npm", "run", "server"]
 EXPOSE 3000
```
Sky11y: Could remove since no use for EXPOSE

backend/src/defaultAvatar.jpg
Sky11y, line 1: The dir where the default img is placed is a bit weird. So was the previous one. Probably some assets dir would be 'best', but let's keep it here, i.e. ignore this opinionated comment.

Sky11y (commented):
Everything related to Avatar is okay. We should probably separate that to it's own branch, since the nginx stuff didn't seem to work properly. There is also another branch called nginx/real-connection-ip that is trying to fix the problem.

backend/src/defaultAvatar.jpg
EyzeCOLD: I agree, but was too lazy to make a specific place for it. I wanted to unweird but went with a half-measure.

backend/dockerfile.dev
EyzeCOLD: donatello

backend/dockerfile
EyzeCOLD: done

backend/nginx/nginx.conf.template
Sky11y, line 1: We could probably set the return status to 429 "Too many requests" with `limit_req-status`.
The default is 503 "Service unavailable"..
Also we should tweak the limits. zone=ws has 5r/m which sound quite restrictive.
I don't know about the others.

Sky11y (commented):
Let's discuss these changes and finish this pr.

Sky11y (approved):
Nice job!

PR #174, opened 2026-05-27
166 sessionconfigts https setting

### LINKED ISSUE #166: sessionConfig.ts https setting
Set the express-session cookie `secure: true` for production.
This prevents cookies to be set with HTTP connection, i.e. HTTPS is required.
Might also be necessary to set the `trust-proxy` variable in express.
RTFM express-session.

Also separate this functionality between dev and prod, since in dev HTTP is used.

Sky11y:
Add `secure: true` option to express-session for production.
Add `NODE_ENV=production` for production build.
Changed production default ports so that this works out of the box in school computers.

The limiter is there still, but will be gone with other PR.
Also please ignore the email related changes which for some reason show as diff, though they are merged to this branch from origin/main.

Closes #166


EvAvKein (approved):
LGTM, optimistic merging, etc

(The prodOverride thing wouldn't work on actual prod, but we're trying to run this on school computers which can't bind to the actual http ports, so 🤷 )

PR #177, opened 2026-05-27
Add multi-tab, same-session handling for file-deleting and for file-repicking

### LINKED ISSUE #150: Editors can get out of sync when mishandled a little
Last tested on May 22.

Recreation steps:

* Open a workspace
* Input some stuff
* Delete the file in another tab for example
* Input some stuff

Frontend gets no notification but backend evicts the doc.

* Wait for a bit but keep the workspace tab open 
* Join the session from another tab with another file (the original was deleted)

Infinite ```Error handling collab request (pushUpdates): RangeError: Applying change set to a document with the wrong length``` on backend.
Frontend gets no info.

* Change the file on one of the tabs

Infinite ```Error handling collab request (pushUpdates): Error: Mismatched change set lengths```.

Doing stuff like this can get the editors out of sync and cause a bunch of weird behavior.

Refreshing the page seems to mostly get stuff to sync up again, but the behavior hardly seems 'optimal'.

EvAvKein:
Title aside, adding a bit more per-character extra time for toasts - this PR has a slightly longer than usual toast and so saw how good this would be.

Closes #150


backend/src/endpoints/collabSocket.ts
Omppu0, line 313:
```suggestion
		if (!fileId) return;
		if (workspace.connectedSockets.values().some((connectedUserId) => connectedUserId == userId)) return;

```

Omppu0: Feels a little hacky but this would fix the issue of sometimes kicking the user out when they have multiple instances of the same workspace open and leave in one of them

Omppu0 (approved):
The stuff seems to work now

